PR #1, opened 2012-12-28
Patricks blob dir perm

jimfulton:
This makes it easier to share blob cache dirs, for example when you want to access the database as a normal user and share the cache with a app server process running as a special user.


tseaver:
@hathawsh Can you figure out if / how Relstorage would be affected?

jamadden:
@jimfulton I don't think RelStorage would be affected. It doesn't use ZODB's `storage_reusable_suite` as far as I can see, it has its own copy. It also doesn't directly use `ZODB.blob.BlobStorage`.

jimfulton:
Thanks. (I was going to figure this out myself, now that I'm practically a RelStorage expert. ;))

I'll freshen this PR soon (unless someone beats me to it, of course).

jimfulton:
Well, the merge with master didn't go well. Still lots of rot.

jimfulton:
Stepping back

This change was pretty conservative. It kept existing behaviour and provided a way to override it.  This was nobley backward compatible.

However, I think everyone (who pays attention) agrees that the existing behavior is stupid.  I propose we simply change the behavior to not specify permissions when creating blob files and directories and let users control this the way they control permissions of other files (e.g. umask)

jimfulton:
Closing in favor of #156

PR #26, opened 2015-01-10
Fix incorrect parameter count on error log

greenmoss:
This fixes the following:

  (stack trace trimmed)
  File "/path/to/virtualenv/lib/python2.7/site-packages/ZODB/Connection.py", line 878, in setstate
    exc_info=sys.exc_info())
TypeError: exception() got an unexpected keyword argument 'exc_info'

### DIFF
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -302,8 +302,8 @@ class Connection(ExportImport, object):
                     f()
                 except: # except what?
                     f = getattr(f, 'im_self', f)
-                    self._log.error("Close callback failed for %s", f,
-                                    exc_info=sys.exc_info())
+                    self._log.error("Close callback failed for %s %s", f,
+                                    sys.exc_info())
             self.__onCloseCallbacks = None
 
         self._debug_info = ()
@@ -873,9 +873,9 @@ class Connection(ExportImport, object):
         except ConflictError:
             raise
         except:
-            self._log.exception("Couldn't load state for %s %s",
+            self._log.exception("Couldn't load state for %s %s %s",
                                 className(obj), oid_repr(oid),
-                                exc_info=sys.exc_info())
+                                sys.exc_info())
             raise
 
     def _setstate(self, obj):


src/ZODB/Connection.py
```diff
@@ -302,8 +302,8 @@ def close(self, primary=True):
                     f()
                 except: # except what?
                     f = getattr(f, 'im_self', f)
-                    self._log.error("Close callback failed for %s", f,
-                                    exc_info=sys.exc_info())
+                    self._log.error("Close callback failed for %s %s", f,
+                                    sys.exc_info())
             self.__onCloseCallbacks = None
```
mgedmin: The original code was correct here.  Your change removes the exception type and traceback from the error log.

src/ZODB/Connection.py
```diff
@@ -873,9 +873,9 @@ def setstate(self, obj):
         except ConflictError:
             raise
         except:
-            self._log.exception("Couldn't load state for %s %s",
+            self._log.exception("Couldn't load state for %s %s %s",
                                 className(obj), oid_repr(oid),
-                                exc_info=sys.exc_info())
+                                sys.exc_info())
```
mgedmin: log.exception(...) is the same as log.error(..., exc_info=True) is the same as log.error(..., exc_info=sys.exc_info()).

The correct fix would be to remove the exc_info argument from the log.exception() call here OR change the .exception() call to an .error() call.

src/ZODB/Connection.py
mgedmin: Apologies, I misspoke: your change doesn't remove the exception type it's part of the 3-tuple returned by exc_info()), but still, the change in the formatting and the loss of the traceback are not good.

mgedmin:
Welp, the Travis build fails due to a recent change in zope.testrunner: https://github.com/zopefoundation/zope.testrunner/commit/3a68ee9c41f4da4f6f64a1fb3fa2801a9a8e322b

src/ZODB/Connection.py
greenmoss: So this is ok in its original form because it's an error, whereas the change below was breaking because it was a warning. Got it.

src/ZODB/Connection.py
greenmoss: OK, How about I change to .error() then?

greenmoss:
It seems my original patches are not correct. Do you want me to close this and start a new pull request, or fix the patches within this pull request?

tseaver:
@greenmoss it is simpler just to fix them and push your changes (keeps the conversation coherent).

jimfulton:
I propose: https://github.com/zopefoundation/ZODB/pull/27 instead.

greenmoss:
Hah! Or that! Closing...

PR #27, opened 2015-01-11
Cleaned up exception logging and changed pins

jimfulton:
- Always use logger.exception rather than logging.error when logging
  exceptions ar the error level.

- Fixed an incorrect call to logging.exceotion (like PR #26, but better :).

- Removed dependency pins and added one to overcome some breakage in
  zope.testrunner.

None of this seems to be CHANGES worthy.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -3,17 +3,18 @@ develop = .
 parts =
     test
     scripts
-versions = versions
 
 [versions]
-zc.recipe.testrunner = 2.0.0
-zc.lockfile = 1.1.0
-zope.interface = 4.0.3
-ZConfig = 3.0.3
-BTrees = 4.0.5
-persistent = 4.0.6
-transaction = 1.4.1
-zdaemon = 4.0.0a1
+# Avoid breakage in 4.4.5:
+zope.testrunner = 4.4.4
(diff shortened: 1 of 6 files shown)


jimfulton:
Damn it, the travis build doesn't use buildout. So my testrunner "fix" didn't work.

Why?

jimfulton:
Attempted travis fix.

jimfulton:
Tests are passing. Can someone review please?

mgedmin:
> None of this seems to be CHANGES worthy.

Fixing a bug actually seen in the wild (#26) seems changelog-worthy to me.

mgedmin:
LGTM.  :shipit:

buildout.cfg
```diff
@@ -3,17 +3,18 @@ develop = .
...
-zc.lockfile = 1.1.0
-zope.interface = 4.0.3
-ZConfig = 3.0.3
-BTrees = 4.0.5
-persistent = 4.0.6
-transaction = 1.4.1
-zdaemon = 4.0.0a1
+# Avoid breakage in 4.4.5:
+zope.testrunner = 4.4.4
+# zc.recipe.testrunner = 2.0.0
```
tseaver: Why leave the others there but commented out?

PR #30, opened 2015-02-25
log ConflictError details, fixes #29

### LINKED ISSUE #29: tryToResolveConflict masks BTreesConflictError details
Most important would be the reason
Including details in ConflictError might be hard, but what about

```
    except (ConflictError, BadClassName), e:
        logger.error(str(e))
```

Might duplicate log entries, but oh well...

agroszer:


tseaver:
LGTM, except for test coverage (we need an assertion that the logging occurs).

mgedmin:
I wish the log message  said something about `"conflict resolution failed with %s: %s" % (e.__class__.__name__, str(e))`.

Perhaps it should be a warning, not an error?  Seeing how the code works anyway.

mgedmin:
Hm.  Correct me if I'm wrong, but, isn't raising ConflictError part of the conflict resolution API, when you want to say "this conflict cannot be resolved"?  E.g. BTree buckets do that in some circumstances.  That's part of normal operation and not an error.  Maybe not even a warning, maybe an info -- "ignoring failed conflict resolution for {class_name} {oid}, the error was: {exc_type}: {exc_value}".

jimfulton:
Marius is correct, conflict errors are not exceptional.  Logging here would be excessively chatty.
Logging as an error demands attention. (Thank you Sentry. :) )

The correct place to log conflict errors normally would be in a client after retries have been exhausted.

I could accept logging at the DEBUG level here.  While conflict errors are normal, a high rate of them effects performance. Note that, at least ZEO, provides metrics to track conflict rates. For debugging, we'd want the full class name of the object in conflict, as well as the exception class name as Marius suggests.

It would also be really nice to include the transaction note to see what was being done at the application level, but that's not handy.  (Conflict-resolution is sooooooo overdue for a re-do.)

jamadden:
@tseaver It looks like this has been updated (cf #34.)

tseaver:
Released via https://pypi.python.org/pypi/ZODB/4.2.0b1

PR #31, opened 2015-04-07
Fix manifest and setup

aclark4life:
- include .coveragerc
- Add author, author_email, keywords

    $ check-manifest
    lists of files in version control and sdist match
    $ pyroma .
    ------------------------------
    Checking .
    Found ZODB
    ------------------------------
    Final rating: 10/10
    Your cheese is so fresh most people think it's a cream: Mascarpone
    ------------------------------


tseaver:
Thanks!

PR #32, opened 2015-04-07
For discussion: Add scripts to create/modify dbs

aclark4life:
- Add fscreate to create db
- Add analyze.py to console scripts as fsanalyze

I'd like to make ZODB to "more approachable to end users" by allowing them to "interact without writing code" e.g.:

```
$ bin/fscreate                                         
Created database with contents {}.
$ bin/fsanalyze Data.fs                                
Processed 1 records in 1 transactions
Average record size is   60.00 bytes
Average transaction size is   60.00 bytes
Types used:
Class Name                                       Count    TBytes    Pct AvgSize
---------------------------------------------- ------- ---------  ----- -------
persistent.mapping.PersistentMapping                 1        60 100.0%   60.00
============================================== ======= =========  ===== =======
                            Total Transactions       1                    0.06k
                                 Total Records       1        0k 100.0%   60.00
                               Current Objects       1        0k 100.0%   60.00
```

Starting with the code in this PR, what additional code/features/etc would be required to get this merged into ZODB? I'm picturing being able to:

```
fscreate 'foo=bar' 
fsadd 'morekeys=morevalues'
```

And so on.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -162,6 +162,8 @@ setup(name="ZODB",
       zip_safe = False,
       entry_points = """
       [console_scripts]
+      fsanalyze = ZODB.scripts.analyze:main
+      fscreate = ZODB.scripts.create:main
       fsdump = ZODB.FileStorage.fsdump:main
       fsoids = ZODB.scripts.fsoids:main
       fsrefs = ZODB.scripts.fsrefs:main
--- a/src/ZODB/scripts/analyze.py
+++ b/src/ZODB/scripts/analyze.py
@@ -144,6 +144,9 @@ def analyze_rec(report, record):
     except Exception as err:
         print(err)
 
-if __name__ == "__main__":
+def main():
     path = sys.argv[1]
     report(analyze(path))
+
+if __name__ == "__main__":
+    main()
--- a/src/ZODB/scripts/create.py
+++ b/src/ZODB/scripts/create.py
@@ -1,0 +1,23 @@
+import ZODB
+import ZODB.FileStorage
+
+
+def main():
+    """
+    Console script to create db
+    """
+    storage = ZODB.FileStorage.FileStorage('Data.fs')
+    db = ZODB.DB(storage)
+    connection = db.open()
+    root = connection.root
+    # XXX Consider allowing end user to create a db w/values
+    # e.g. createdb 'foo=bar'
+    print('Created database with contents %s.' % root._root)
+
+
+if __name__ == '__main__':
+
+    """
+    Execute console script to create db
+    """
+    main()


jimfulton:
On Tue, Apr 7, 2015 at 10:26 AM, Alex Clark <notifications@github.com>
wrote:

>
>    - Add fscreate to create db
>    - Add analyze.py to console scripts as fsanalyze
>
> I'd like to make ZODB to "more approachable to end users"
>
Great!


by allowing them to "interact without writing code" e.g.:
>
I'm unsure why this would make ZODB more approachable.




> $ bin/fscreate
> Created database with contents {}.
> $ bin/fsanalyze Data.fs
> Processed 1 records in 1 transactions
> Average record size is   60.00 bytes
> Average transaction size is   60.00 bytes
> Types used:
> Class Name                                       Count    TBytes    Pct AvgSize
> ---------------------------------------------- ------- ---------  ----- -------
> persistent.mapping.PersistentMapping                 1        60 100.0%   60.00
> ============================================== ======= =========  ===== =======
>                             Total Transactions       1                    0.06k
>                                  Total Records       1        0k 100.0%   60.00
>                                Current Objects       1        0k 100.0%   60.00
>
>
I don't understand why this would make ZODB more approachable.


>  Starting with the code in this PR, what additional code/features/etc
> would be required to get this merged into ZODB? I'm picturing being able to:
>
> fscreate 'foo=bar'
> fsadd 'morekeys=morevalues'
>
> And so on.
>
I don't understand what these would do.

I'm grateful that you want to help and make ZODB more approachable.

I have some problems with this, um, approach:

I don't think a bunch of command-line scripts are more approachable than a
Python
prompt. The command-line is a constricted form form of programming.

I'd *much* rather make the Python prompt experience better.

In recent years, I've made small steps, like making it easier to
construct database objects::

    conn = ZODB.DB('my.fs')

There are a number of other small steps that could improve things, like
maybe allowing simpler iteration:

    for trans in db:
       for record in trans:
          ...

As far as exploring database usage, making easy for people to use
iterators would go a long way.

Another thing that I think would help would be making it easy to create
collections
(ie sets) of persistent objects.  I think I got 80% of the way to this, but
got bogged down
in the Python implementation. :(

In any case, I think efforts to make ZODB more approachable should be
documentation driven.
I recommend coming up with compelling science fiction and then making it
come true.
That's how I decided we needed PxBTrees.

I'd be very happy to sprint on this some time. Many people interested in
ZODB
happen to be in the DC area (for some definition of the DC area. :)

Jim



-- 
Jim Fulton
http://www.linkedin.com/in/jimfulton

aclark4life:
@jimfulton It makes it more approachable to me because if I:

```
$ pip install ZODB
$ ls bin
```

I expect to see something called ``zodb`` that does something useful e.g. ``zodbcreate``. In lieu of that and following existing ZODB conventions, I created ``fscreate``. So now when I see ``fscreate`` in ``./bin/``, I think "OK this ZODB package probably does something useful." Actually, I had that thought when I saw ``fsrefs``, etc. but then I thought "OK I can execute these scripts once I have a database, but I have to write a script to get a database first. Annoying." If I don't see any scripts, I think "this is a library".

``analyze`` is an existing script with no corresponding console script. I liked its output so I added a console script for it. Being able to create and analyze a database within seconds of installing it reinforces the idea that "this ZODB package does something immediately useful".

The next most useful thing would be to create a database with values in it e.g. 

```
fscreate "{'name': 'Alex Clark'}"
```

And the next most useful thing after that would be to add entries to an existing database e.g.

```
fsadd Data.fs "{'name': 'Alex Clark'}"
```

And so on. These don't have to go out in the ZODB package, but that's my first stop.

navytux:
For the reference: [zodbtools](https://pypi.org/project/zodbtools/) provides `zodb` command with various subcommands to work on a ZODB database like `info`, `dump`, `commit`, `analyze`, etc. It is kept separate from ZODB because @jimfulton [adviced](https://github.com/zopefoundation/ZODB/pull/128#issuecomment-260970932) us not to load ZODB with scripts anymore.

PR #22, opened 2014-06-04
Fix #21, FileStorage: when disk space runs out while packing, do not lea...

agroszer:
...ve the ``.pack`` file around

PR #34, opened 2015-05-19
Support PyPy. 

### LINKED ISSUE #33: Explicit PyPy Support (in progress)
ZODB currently does not cleanly pass a test run under PyPy --- there are numerous failures.

Part of this is due to the lack of `noload` in PyPy's "cPickle" module (which is actually implemented in Python) --- this is easily solved through the use of `zodbpickle` (which is already a dependency). At least part of the rest of it is due to differences in the C and pure-Python implementations of `persistent` (e.g., the Python `PickleCache` lacks some size attributes found in the C version). 

Hopefully the rest are minor differences, but I'm not sure yet. I'm working on a branch at https://github.com/NextThought/ZODB/tree/pypy to find out, and hopefully fix them (looks like some other PRs in `persistent` will be useful, too).

I'm opening this issue to let people know someone is working on this, and maybe to have a place to ask for help if I get stuck :)

jamadden:
Fixes #33. Depends on persistent 4.1.0. Also supports jython, though you have to run the tests manually so that's not mentioned.

Highlights of changes:

- Use zodbpickle under not just PyPy but Python2.7. This fixes the broken `noload` problem that's encountered by many of the tools and things like relstorage and zc.zodbdgc.
- Centralize the creation of picklers and unpicklers, smoothing over the differences between Python versions and zodbpickle. This takes care of an existing TODO.
- The Blob class needs to use an external weakref, not one stored on `self`. See https://bitbucket.org/pypy/pypy/issue/2030/circular-weakref-callbacks-never-called

Other than that, the changes are pretty minor and mostly in doctests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,6 @@ Programming Language :: Python :: 3.3
 Programming Language :: Python :: 3.4
 Programming Language :: Python :: Implementation :: CPython
 Programming Language :: Python :: Implementation :: PyPy
-Programming Language :: Python :: Implementation :: Jython
 Topic :: Database
 Topic :: Software Development :: Libraries :: Python Modules
 Operating System :: Microsoft :: Windows
--- a/setup.py
+++ b/setup.py
@@ -60,6 +60,7 @@ Programming Language :: Python :: 3.3
 Programming Language :: Python :: 3.4
 Programming Language :: Python :: Implementation :: CPython
 Programming Language :: Python :: Implementation :: PyPy
+Programming Language :: Python :: Implementation :: Jython
 Topic :: Database
 Topic :: Software Development :: Libraries :: Python Modules
 Operating System :: Microsoft :: Windows
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ to application logic.  ZODB includes features such as a plugable storage
 interface, rich transaction support, and undo.
 """
 
-VERSION = "4.1.0"
+VERSION = "4.2.0.dev0"
 
 import os
 import platform
@@ -59,6 +59,7 @@ Programming Language :: Python :: 3.2
 Programming Language :: Python :: 3.3
 Programming Language :: Python :: 3.4
 Programming Language :: Python :: Implementation :: CPython
+Programming Language :: Python :: Implementation :: PyPy
 Topic :: Database
 Topic :: Software Development :: Libraries :: Python Modules
 Operating System :: Microsoft :: Windows
@@ -158,7 +159,7 @@ setup(name="ZODB",
       tests_require = tests_require,
       extras_require = dict(test=tests_require),
       install_requires = [
-        'persistent',
+        'persistent', # XXX: When new persistent release is out need to add version number for PyPy
         'BTrees >= 4.1.2',
         'ZConfig',
         'transaction >= 1.4.1' if PY3 else 'transaction',
--- a/setup.py
+++ b/setup.py
@@ -160,10 +160,10 @@ setup(name="ZODB",
       tests_require = tests_require,
       extras_require = dict(test=tests_require),
       install_requires = [
-        'persistent', # XXX: When new persistent release is out need to add version number for PyPy
-        'BTrees >= 4.1.2',
+        'persistent >= 4.1.0',
+        'BTrees >= 4.1.3',
         'ZConfig',
-        'transaction >= 1.4.1' if PY3 else 'transaction',
+        'transaction >= 1.4.4',
         'six',
         'zc.lockfile',
         'zdaemon >= 4.0.0a1',
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,7 @@ interface, rich transaction support, and undo.
 VERSION = "4.1.0"
 
 import os
+import platform
 import sys
 from setuptools import setup, find_packages
 
@@ -35,6 +36,10 @@ if (3,) < sys.version_info < (3, 2):
     sys.exit(0)
 
 PY3 = sys.version_info >= (3,)
+PY27 = sys.version_info >= (2,7)
+py_impl = getattr(platform, 'python_implementation', lambda: None)
+PYPY = py_impl() == 'PyPy'
+
 
 # The (non-obvious!) choices for the Trove Development Status line:
 # Development Status :: 5 - Production/Stable
@@ -154,14 +159,14 @@ setup(name="ZODB",
       extras_require = dict(test=tests_require),
       install_requires = [
         'persistent',
-        'BTrees',
+        'BTrees >= 4.1.2',
         'ZConfig',
         'transaction >= 1.4.1' if PY3 else 'transaction',
         'six',
         'zc.lockfile',
         'zdaemon >= 4.0.0a1',
         'zope.interface',
-        ] + (['zodbpickle >= 0.2'] if PY3 else []),
+        ] + (['zodbpickle >= 0.6.0'] if (PY3 or PY27 or PYPY) else []),
       zip_safe = False,
       entry_points = """
       [console_scripts]
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -439,7 +439,6 @@ class Connection(ExportImport, object):
         # the savepoint, then they won't have _p_oid or _p_jar after
         # they've been unadded. This will make the code in _abort
         # confused.
-
         self._abort()
 
         if self._savepoint_storage is not None:
@@ -463,7 +462,6 @@ class Connection(ExportImport, object):
                 if obj._p_changed:
                     obj._p_changed = False
             else:
-
                 # Note: If we invalidate a non-ghostifiable object
                 # (i.e. a persistent class), the object will
                 # immediately reread its state.  That means that the
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -1018,10 +1018,9 @@ class Connection(ExportImport, object):
         items = self._cache.lru_items()
         # fine everything. some on the lru list, some not
         everything = self._cache.cache_data
-        # remove those items that are on the lru list (which may not actually
-        # be in the full cache, under the Python implementation)
+        # remove those items that are on the lru list
         for k,v in items:
-            everything.pop(k, None)
+            del everything[k]
         # return a list of [ghosts....not recently used.....recently used]
         return list(everything.items()) + items
 
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -328,10 +328,10 @@ class Connection(ExportImport, object):
                 # get back here.
         else:
             self.opened = None
-            
+
         am = self._db._activity_monitor
         if am is not None:
-            am.closedConnection(self)        
+            am.closedConnection(self)
 
     def db(self):
         """Returns a handle to the database this connection belongs to."""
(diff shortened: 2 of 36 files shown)


jamadden:
I also have some minor changes to ZEO for PyPy support, but because ZEO re-uses many of the test cases from ZODB with its own storage, I can't really open that PR until this one is merged and released because the Travis tests will fail.

jamadden:
PyPy3 failed because .travis.yml pins zope.testrunner to 4.4.4, but 4.4.7 is required to workaround a bug in PyPy3 I didn't move the pin because I'm not sure why it's there. Can it be updated?

mgedmin:
@jimfulton [didn't](https://github.com/zopefoundation/ZODB/commit/6f64c2aa44271a0efd2bd2f3c07401a4253d1a0a) [explain](https://github.com/zopefoundation/ZODB/commit/1bbaa6e270a4f13f4dc60e6c0a67b940d50fc060) the reason for the zope.testrunner pin, but I'm guessing it's because of https://github.com/zopefoundation/zope.testrunner/issues/28.  IOW tests run faster with zope.testrunner 4.4.4 than with 4.4.5--4.4.8.

There's an outstanding PR for zope.testrunner that fixes the slowdown.  I was planning to merge it today and release 4.4.9.  I say: remove the pin.

setup.py
```diff
@@ -153,15 +159,15 @@ def read_file(*path):
...
         'ZConfig',
-        'transaction >= 1.4.1' if PY3 else 'transaction',
+        'transaction >= 1.4.4',
         'six',
         'zc.lockfile',
         'zdaemon >= 4.0.0a1',
         'zope.interface',
-        ] + (['zodbpickle >= 0.2'] if PY3 else []),
+        ] + (['zodbpickle >= 0.6.0'] if (PY3 or PY27 or PYPY) else []),
       zip_safe = False,
```
tseaver: Does using `zodbpickle` on Python 2.7 introduce data interchange issues, e.g. with clients on 2.6, or older ZODB versions?

src/ZODB/Connection.py
```diff
@@ -1020,9 +1018,10 @@ def _cache_items(self):
         items = self._cache.lru_items()
         # fine everything. some on the lru list, some not
         everything = self._cache.cache_data
-        # remove those items that are on the lru list
+        # remove those items that are on the lru list (which may not actually
+        # be in the full cache, under the Python implementation)
```
tseaver: Is this change stlil true (i.e., can the Python implementation of picklecache return items from `lru_items` which are not in the cache?)

src/ZODB/DB.py
```diff
@@ -555,12 +559,15 @@ def f(con, detail=detail, rc=sys.getrefcount, conn_no=conn_no):
                 # sys.getrefcount(ob) returns.  But, in addition to that,
                 # the cache holds an extra reference on non-ghost objects,
                 # and we also want to pretend that doesn't exist.
+                # If we have no way to get a refcount, we return False to symbolize
+                # that. As opposed to None, this has the advantage of being usable
+                # as a number (0) in case clients depended on that.
```
tseaver: Wouldn't it be better just to return `0`?

src/ZODB/DemoStorage.test
```diff
@@ -375,12 +385,12 @@ Now, we create a demostorage.
 
 If we ask for an oid, we'll get 1042.
 
-    >>> u64(storage.new_oid())
+    >>> print(u64(storage.new_oid()))
```
tseaver: Why?

src/ZODB/_compat.py
```diff
@@ -11,15 +11,26 @@
...
 try:
     # Python 2.x
-    from cPickle import Pickler
-    from cPickle import Unpickler
-    from cPickle import dump
-    from cPickle import dumps
-    from cPickle import loads
-    from cPickle import HIGHEST_PROTOCOL
+    import cPickle
+    if (hasattr(cPickle.Unpickler, 'load') and not hasattr(cPickle.Unpickler, 'noload')) or sys.version_info >= (2,7):
```
tseaver: Wrap at 80 characters,

src/ZODB/_compat.py
```diff
@@ -61,14 +72,58 @@ def loads(s):
...
+    p = Pickler(*args, **kwargs)
+    if sys.version_info[0] < 3:
+        p.inst_persistent_id = persistent_id
+        # PyPy uses a python implementation of cPickle in both Python 2
+        # and Python 3. We can't really detect inst_persistent_id as its
+        # a magic attribute that's not readable, but it doesn't hurt to
+        # simply always assign to persistent_id also
+        p.persistent_id = persistent_id
+    else:
+        p.persistent_id = persistent_id
```
tseaver: Clearer to move the assignment / comment outside the `if` altogether.

setup.py
jamadden: Good question. As far as I can see, no, it shouldn't. The default protocol didn't change, and zodbpickle is based on the pickle modules from Python 2.7 anyway. The fancy string encoding stuff zodbpickle can do is only used under Python 3.

src/ZODB/_compat.py
```diff
@@ -61,14 +72,58 @@ def loads(s):
...
-    from cStringIO import StringIO as BytesIO
+    if IS_JYTHON:
+        # Jython 2.7rc2 cStringIO.StringIO class has a bug
+        # resulting in StringIndexOutOfBoundExceptions
+        # when repeatedly writing and then seeking back to 0
+        # http://bugs.jython.org/issue2324
+        from io import BytesIO
+    else:
+        # XXX: why not just import BytesIO from io?
+        from cStringIO import StringIO as BytesIO
```
tseaver: Given that the fix for http://bugs.jython.org/issue2324 [landed in Jython 2.7.0](https://hg.python.org/jython/file/412a8f9445f7/NEWS#l17), can we just back this part out??

Also, I'm still not clear why we are not just using `io.BytesIO` everywhere.

src/ZODB/Connection.py
jamadden: Ah. No, it doesn't seem to be anymore. At least, the tests that failed before that change aren't failing if I revert it. If it does fail, it means that the picklecache is corrupted, and we have a CFFI level memory leak (under PyPy).

src/ZODB/tests/testRecover.py
```diff
@@ -69,12 +69,28 @@ def populate(self):
     def damage(self, num, size):
         self.storage.close()
         # Drop size null bytes into num random spots.
-        for i in range(num):
+        for i in range(num - 1):
```
tseaver: Why?

src/ZODB/DB.py
jamadden: This is used in a diagnostic method, which is usually, but not always, printed. It seemed useful to me to be able to distinguish the "can't get a refcount" case from the "nobody but the cache has this item" case.

src/ZODB/tests/testSerialize.py
```diff
@@ -139,7 +164,17 @@ def test_funky_datetime_serialization(self):
...
         environ = os.environ.copy()
-        environ['PYTHONPATH'] = os.pathsep.join(sys.path)
+        if IS_JYTHON:
+            # Jython 2.7rc2 has a bug; if its Lib directory is
+            # specifically put on the PYTHONPATH, then it doesn't add
+            # it itself, which means it fails to 'import site' because
+            # it can't import '_jythonlib' and the whole process fails
+            # We would use multiprocessing here, but it doesn't exist on jython
+            sys_path = [x for x in sys.path
+                        if not x.endswith('Lib') and x != '__classpath__' and x!= '__pyclasspath__/']
```
tseaver: AIYAH!

src/ZODB/tests/test_fsdump.py
```diff
@@ -58,7 +58,7 @@
 Trans #00001 tid=... time=... offset=<OFFSET>
     status=' ' user='' description='added an OOBTree'
   data #00000 oid=0000000000000000 size=<SIZE> class=persistent.mapping.PersistentMapping
-  data #00001 oid=0000000000000001 size=<SIZE> class=BTrees.OOBTree.OOBTree
+  data #00001 oid=0000000000000001 size=<SIZE> class=BTrees.OOBTree.OOBTree...
```
tseaver: What are we eliding here?

src/ZODB/_compat.py
jamadden: > Given that the fix for http://bugs.jython.org/issue2324 landed in Jython 2.7.0, can we just back this part out??

Yes. I should have gardened that. Sorry.

> Also, I'm still not clear why we are not just using io.BytesIO everywhere.

Don't look at me. I just preserved existing behaviour :)

tox.ini
```diff
@@ -29,6 +32,10 @@ commands =
     python setup.py test -q
 deps = {[testenv]deps}
 
+[testenv:jython]
```
tseaver: This stanza is not needed: tox already knows how to invoke jython (even if it can't build the virtualenv ATM).

src/ZODB/DemoStorage.test
jamadden: Because `u64` returns a long-integer, and one of the platforms (probably jython) returned `1042L` instead of `1042`, thus breaking the doctest. Using `print` causes `__str__` to be called instead of `__repr__` and gets consistent results everywhere.

src/ZODB/_compat.py
jamadden: Will do. I just moved existing code here from `ZODB.serialize`, but it's code that I originally wrote :)

src/ZODB/tests/test_fsdump.py
jamadden: Under CPython, class is as you see it. Under PyPy or PURE_PYTHON, class is `OOBTreePy`.

src/ZODB/tests/testSerialize.py
jamadden: Don't think that wasn't fun to track down. The jython guys are aware of some problems with import and have recently landed an overhaul, although I haven't had a chance to test it.

src/ZODB/DemoStorage.test
tseaver: OK

src/ZODB/tests/test_fsdump.py
tseaver: OK

src/ZODB/tests/testSerialize.py
tseaver: A link to the Jython bug for the issue would be helpful (e.g., so that we could remove the cruft at some point).

src/ZODB/DB.py
tseaver: OK

tseaver:
Thank you!

tseaver:
Memory fails me:  does the ZEO PR need this to be released, or does it use a clone of ZODB for testing?

jamadden:
ZEO is something of the odd one out here. It uses tox.ini in the same way ZODB does for local testing, but on Travis it sets up a full buildout environment. As far as I can see from the `buildout.cfg` and `setup.py`, there's nothing special there to clone ZODB, so it would need a release.

src/ZODB/tests/testSerialize.py
jamadden: I just tried to create a small test case to file a bug report with, and I can't reproduce this problem in Jython 2.7.0. It's definitely there in 2.7rc2, although I don't see any specific release notes about fixing something like this for 2.7.0. It's possible it was in the meta-bug [2319](http://bugs.jython.org/issue2319) which was fixed for rc3.

tseaver:
@agroszer I'm hoping to get an updated PR #22 landed today, before making a release.  And maybe an updated version of #30, logging at `DEBUG` level.  Do you have headspace to tackle them soonish?

agroszer:
@tseaver Today definitely not, soonish yes.

jamadden:
@tseaver The updates needed for #22 and #30 look relatively minor, I can take a pass at updating them, if @agroszer doesn't mind.

tseaver:
Released via https://pypi.python.org/pypi/ZODB/4.2.0b1

PR #39, opened 2015-06-15
Add a new option to recover only to a certain point in time.

stephan-hof:
Hi guys,

we had a case were multiple (very bad) changes on different places inside ZODB have been performed.
The only thing we had known was "the date where it was in a good/consistent" state.
Hence I enhanced fsrecover.py to recover transactions only to this point and time and omit all the others.


jimfulton:
Can you write a test for this?

PR #37, opened 2015-05-25
Use PEP-426 environment markers

mgedmin:
Should fix #36.

Issue #36, opened 2015-05-25
dynamic install_requires breaks universal wheels


mgedmin:
pip 7 builds and caches wheels, which speeds up repeated installs.

Unfortunately, if you build a wheel of ZODB on Python 2.6, it will [compute](https://github.com/zopefoundation/ZODB/blob/master/setup.py#L170) `install_requires` that works on 2.6 but fails on 2.7 (because no 'zodbpickle').  This wheel will be named `ZODB-4.2.0b1-py2-none-any.whl`, and so pip will try to use it on 2.7. This [breaks](https://jenkins.gedmin.as/view/Failures/job/zodbbrowser-on-windows/183/console) people's buildbots.

STEPS TO REPRODUCE:

```
$ virtualenv --version
13.0.1

$ cat tox.ini
[tox]
envlist = py26,py27
skipsdist = true

[testenv]
skip_install = true
deps = ZODB
commands =
  python -c 'import ZODB'

$ rm ~/.cache/pip/wheels/*/*/*/*/ZODB-*.whl

$ tox --pre
py26 installed: argparse==1.3.0,BTrees==4.1.3,persistent==4.1.0,six==1.9.0,transaction==1.4.4,wheel==0.24.0,zc.lockfile==1.1.0,ZConfig==3.0.4,zdaemon==4.1.0,ZODB==4.2.0b1,zope.interface==4.1.2
py26 runtests: PYTHONHASHSEED='2859500558'
py26 runtests: commands[0] | python -c import ZODB
py27 recreate: /tmp/bork/.tox/py27
py27 installdeps: ZODB
py27 installed: BTrees==4.1.3,persistent==4.1.0,six==1.9.0,transaction==1.4.4,wheel==0.24.0,zc.lockfile==1.1.0,ZConfig==3.0.4,zdaemon==4.1.0,ZODB==4.2.0b1,zope.interface==4.1.2
py27 runtests: PYTHONHASHSEED='2859500558'
py27 runtests: commands[0] | python -c import ZODB
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/tmp/bork/.tox/py27/local/lib/python2.7/site-packages/ZODB/__init__.py", line 28, in <module>
    from ZODB.DB import DB, connection
  File "/tmp/bork/.tox/py27/local/lib/python2.7/site-packages/ZODB/DB.py", line 23, in <module>
    from ZODB.broken import find_global
[...]

PR #42, opened 2015-11-03
Make zodbpickle non-optional. 

jamadden:
As [discussed on the mailing list](https://groups.google.com/forum/#!topic/zodb/dRP-AGS7hmQ). Ref #36 and #37 

The only potential compatibility issue I'm aware of for Python 2.6 users is the [change in exceptions that are raised](https://github.com/zopefoundation/ZODB/issues/36#issuecomment-108438240). I wasn't sure if/where to mention that. 

Arguably this should be a bigger version bump than just 4.2.1 but I didn't change that either.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
     - 3.3
     - 3.4
 install:
-    - travis_retry pip install BTrees ZConfig manuel persistent six transaction zc.lockfile zdaemon zope.interface zope.testing zope.testrunner
+    - travis_retry pip install BTrees ZConfig manuel persistent six transaction zc.lockfile zdaemon zodbpickle zope.interface zope.testing zope.testrunner
     - travis_retry pip install -e .
 script:
     - zope-testrunner -u --test-path=src --auto-color --auto-progress
--- a/src/ZODB/_compat.py
+++ b/src/ZODB/_compat.py
@@ -12,20 +12,20 @@
 #
 ##############################################################################
 import sys
+from six import PY3
 
 IS_JYTHON = sys.platform.startswith('java')
 
-try:
+
+if not PY3:
     # Python 2.x
-    import cPickle
-    if ((hasattr(cPickle.Unpickler, 'load') and not hasattr(cPickle.Unpickler, 'noload')) or
-		sys.version_info >= (2,7)):
-        # PyPy doesn't have noload, and noload is broken in Python 2.7.
-        # Get the fastest version we can (PyPy has no fastpickle)
-        try:
-            import zodbpickle.fastpickle as cPickle
-        except ImportError:
-            import zodbpickle.pickle as cPickle
+    # PyPy's cPickle doesn't have noload, and noload is broken in Python 2.7,
+    # so we need zodbpickle.
+    # Get the fastest working version we can (PyPy has no fastpickle)
+    try:
+        import zodbpickle.fastpickle as cPickle
+    except ImportError:
+        import zodbpickle.pickle as cPickle
     Pickler = cPickle.Pickler
     Unpickler = cPickle.Unpickler
     dump = cPickle.dump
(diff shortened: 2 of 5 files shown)


tox.ini
```diff
@@ -20,6 +20,7 @@ deps =
     transaction
     zc.lockfile
     zdaemon
+    zodbpickle
```
mgedmin: Technically, you only need to list packages from `tests_require` here; things that are in `install_requires` will already be installed by tox.  (There might be one or two exceptions, e.g. if we `install_requires=[..., 'foo', ...]` and `foo` has a `setup_requires` on `bar`, and `bar` isn't already installed explicitly, we might get a `./.eggs/bar-*.egg` that is prevented if we explicitly list `bar` in deps here.)

mgedmin:
LGTM.

jamadden:
Except I broke Python 3 (I only ran tox on 2.6 locally) :( Fix forthcoming.

tox.ini
jamadden: The deps block is proceeded by this comment, so I assume somebody had a reason for listing all the deps; I just followed along:

```
# without explicit deps, setup.py test will download a bunch of eggs into $PWD
deps =
    BTrees
    ...
```

.travis.yml
```diff
@@ -9,7 +9,7 @@ python:
     - 3.3
     - 3.4
 install:
-    - travis_retry pip install BTrees ZConfig manuel persistent six transaction zc.lockfile zdaemon zope.interface zope.testing zope.testrunner
+    - travis_retry pip install BTrees ZConfig manuel persistent six transaction zc.lockfile zdaemon zodbpickle zope.interface zope.testing zope.testrunner
```
mgedmin: This seems wrong: `pip install -e .` should be pulling in zodbpickle

tox.ini
mgedmin: I believe I was the one who added that comment originally.  ;-)

Not necessarily in this particular tox.ini; things like that tend to be cargo-culted around.

.travis.yml
jamadden: I agree that's a lot of repetition there, but it fits the pattern of what was done before. Perhaps we can clean up .travis.yml in a subsequent PR (enable caching and so forth)?

mgedmin: +1

jamadden:
Build's green, python 3 fixed. (Hmm, we should probably add 3.5 to the matrix at some point, probably after [BTrees work on 3.5](https://github.com/zopefoundation/BTrees/pull/16))

mgedmin:
Thank you!

jamadden:
Would anyone else care to review, maybe @tseaver who was involved in the mailing list conversation?

src/ZODB/_compat.py
```diff
@@ -12,20 +12,20 @@
 #
 ##############################################################################
 import sys
+from six import PY3
 
 IS_JYTHON = sys.platform.startswith('java')
 
-try:
+
+if not PY3:
```
tseaver: This feels wrong:  for instance, what about PyPy3?

jamadden: The other half of the branch (the "if PY3" case) handles PyPy3. There was already a special case for any Python 3 version (they all used zodbpickle already), but it was being detected implicitly through the failure of `import cPickle`, which never existed on any Python 3 (but did exist on PyPy2, albeit the same as regular `pickle`).

jamadden: I can use `if PY2` if that seems cleaner. I just used `not PY3` to try to keep the sense of the test the same as the try/except version.

agroszer:
please release this ASAP, drives me nuts

mgedmin:
Oh, right, I don't have PyPI rights to make ZODB releases.

Sorry, @agroszer, you'll have to wait some more.

PR #43, opened 2015-11-07
Cleanup tox.ini and .travis.yml as mentioned in #42 (and Python 3.5)

jamadden:
- (Both) Remove repetition of deps. Only install tests_require deps
  explicitly. The previous comment in tox.ini about eggs in the
  working directory doesn't apply, per @mgedmin.
- Setup Travis to use the local wheel cache for faster install times. I stole this from what I do in gevent, which works, but may not be optimal.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ python:
     - 3.2
     - 3.3
     - 3.4
+    - 3.5
 install:
     # First install a newer pip so that it can use the wheel cache
     # (only needed until travis upgrades pip to 7.x; note that the 3.5
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,25 @@ python:
     - 3.3
     - 3.4
 install:
-    - travis_retry pip install BTrees ZConfig manuel persistent six transaction zc.lockfile zdaemon zodbpickle zope.interface zope.testing zope.testrunner
-    - travis_retry pip install -e .
+    # First install a newer pip so that it can use the wheel cache
+    # (only needed until travis upgrades pip to 7.x; note that the 3.5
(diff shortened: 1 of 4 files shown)


jamadden:
The use of Travis dep caching saves ~25 seconds for the CPython builds (due to not rebuilding the C extensions for BTrees and persistent), taking the total test time down from 4:30 to ~4. Not a really huge gain (on gevent it saved a minute or more because Cython builds compiles itself) but not nothing either. I haven't had any problems with it on gevent so I don't predict it becoming a burden.

jamadden:
Any thoughts? Esp @mgedmin ? Your cargo cult eagerly awaits :)

mgedmin:
Travis caching is a brilliant thing.

For a couple of packages I've replaced the explicit dependency lists in tox.ini with `.[test]`, and later noticed that this causes `.eggs/*.egg` to be created, because of `setup_requires` :-(  In some cases this is unavoidable, even with all the dependencies explicitly listed (and topologically sorted), because AFAIU tox runs setup.py egg_info/sdist of the main package using basepython, which means if the main package uses setup_requires, those deps will be downloaded anyway.  In other cases listing the dependencies in tox.ini can avoid the .eggs/ garbage.  So I just don't know.  Thinking about this issue makes me want to go away and do something else instead.

jamadden:
Given the newly [3.5 compatible release of BTrees](zopefoundation/BTrees#17), this seems like a good time to add 3.5 to the testing matrix and claim support for it.

jamadden:
Any objections to merging this?

mgedmin:
LGTM, and travis doesn't object either.

Issue #45, opened 2016-03-16
Uncaught exception when shutting down server


do3cc:
Copy&Pasted from https://github.com/zodb/relstorage/issues/14
written by @neilferreira

Hi all

I'm getting an error when shutting down my server in my development environment.
Plone 4.3.3
RelStorage = 1.5.1

sauna and get the same issue when CTRL + C'ing the insance.

```
[instance]
<= instance_base
recipe = plone.recipe.zope2instance
http-address = 8080
blob-dir ${buildout:var-dir}/blobstorage
dsn dbname = ${buildout:database-name} host=${buildout:database-host} user=${buildout:database-user}

^C2014-06-17 14:15:21 INFO SignalHandler Caught signal SIGINT
2014-06-17 14:15:21 INFO Z2 Shutting down
Traceback (most recent call last):
  File "/home/neilf/development/website/website-website.wa.gov.au-v2/parts/instance/bin/interpreter", line 328, in 
    exec(compile(file__f.read(), __file, "exec"))
  File "/home/neilf/development/website/buildout-cache/eggs/Zope2-2.13.22-py2.7.egg/Zope2/Startup/run.py", line 76, in 
    run()
  File "/home/neilf/development/website/buildout-cache/eggs/Zope2-2.13.22-py2.7.egg/Zope2/Startup/run.py", line 26, in run
    starter.run()
  File "/home/neilf/development/website/buildout-cache/eggs/Zope2-2.13.22-py2.7.egg/Zope2/Startup/init.py", line 108, in run
    self.shutdown()
  File "/home/neilf/development/website/buildout-cache/eggs/Zope2-2.13.22-py2.7.egg/Zope2/Startup/init.py", line 113, in shutdown
    db.close()
  File "/home/neilf/development/website/buildout-cache/eggs/ZODB3-3.10.5-py2.7-linux-x86_64.egg/ZODB/DB.py", line 624, in close
    @self._connectionMap
  File "/home/neilf/development/website/buildout-cache/eggs/ZODB3-3.10.5-py2.7-linux-x86_64.egg/ZODB/DB.py", line 506, in _connectionMap
    self.pool.map(f)
  File "/home/neilf/development/website/buildout-cache/eggs/ZODB3-3.10.5-py2.7-linux-x86_64.egg/ZODB/DB.py", line 206, in map
    self.all.map(f)
[...]

PR #46, opened 2016-03-16
Fixing uncaught exception problem on shutdown

do3cc:
Nested connections try to clean up themselves multiple
times. This does not work.

I'd like to have a review from somebody, I don't feel 1000 % confident about the change.

### DIFF
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -18,7 +18,9 @@ Bugs Fixed
 - Pinned the ``transaction`` and ``manuel`` dependencies to Python 2.5-
   compatible versions when installing under Python 2.5.
 
-- Cleanup of nested mvcc connections could have failed.
+- Avoid failure during cleanup of nested databases that provide MVCC
+  on storage level (Relstorage).
+  https://github.com/zopefoundation/ZODB/issues/45
 
 3.10.5 (2011-11-19)
 ===================
--- a/src/CHANGES.txt
+++ b/src/CHANGES.txt
@@ -18,6 +18,8 @@ Bugs Fixed
 - Pinned the ``transaction`` and ``manuel`` dependencies to Python 2.5-
   compatible versions when installing under Python 2.5.
 
+- Cleanup of nested mvcc connections could have failed.
+
 3.10.5 (2011-11-19)
 ===================
 
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -1073,7 +1073,8 @@ class Connection(ExportImport, object):
     def _release_resources(self):
         for c in self.connections.itervalues():
             if c._mvcc_storage:
-                c._storage.release()
+                if c._storage is not None:
+                    c._storage.release()
             c._storage = c._normal_storage = None
             c._cache = PickleCache(self, 0, 0)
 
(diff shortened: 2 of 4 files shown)


do3cc:
#45

src/CHANGES.txt
```diff
@@ -18,6 +18,8 @@ Bugs Fixed
 - Pinned the ``transaction`` and ``manuel`` dependencies to Python 2.5-
   compatible versions when installing under Python 2.5.
 
+- Cleanup of nested mvcc connections could have failed.
```
tseaver: The convention is to have the changelog entries in imperative mode.  Something like:

```
- Avoid failure due to repeated cleanup of a connection's storage (typically during shutdown).
```

tseaver:
Aside from the changelog issue, LGTM.

mgedmin:
How hard would it be to write a unit test for this?

do3cc:
@mgedmin very hard. There comes the I am not 1000% sure what I am doing into play. But I am trying.

src/ZODB/tests/testMVCCMappingStorage.py
```diff
@@ -33,6 +33,15 @@
...
 
 class MVCCTests:
+    def checkClosingNestedDatabasesWorks(self):
+        # This tests for the error described in
+        # https://github.com/zopefoundation/ZODB/issues/45
+        db1 = DB(self._storage)
+        db2 = DB(None, databases=db1.databases, database_name='2')
+        db1.open().get_connection('2')
+        db1.close()
+        db2.close()
```
do3cc: Zope stores all dbs that have been created. On shutdown, it closes each of them, I am mirroring this here.

src/ZODB/tests/testMVCCMappingStorage.py
```diff
@@ -33,6 +33,15 @@
     )
 
 class MVCCTests:
+    def checkClosingNestedDatabasesWorks(self):
+        # This tests for the error described in
+        # https://github.com/zopefoundation/ZODB/issues/45
+        db1 = DB(self._storage)
+        db2 = DB(None, databases=db1.databases, database_name='2')
+        db1.open().get_connection('2')
```
do3cc: I must admin, I am only triggering observed behavior. From my understanding, the information that db1 has somehow a connection to db2 is lazy added when I try to get the second connection from it.

src/ZODB/Connection.py
```diff
@@ -1073,7 +1073,8 @@ def _resetCache(self):
     def _release_resources(self):
         for c in self.connections.itervalues():
             if c._mvcc_storage:
-                c._storage.release()
+                if c._storage is not None:
+                    c._storage.release()
```
do3cc: When I wrote the fix, I compared a stock Plone page with our page with Relstorage.
The reason why this causes no exception in a stock Plone/Zope page is that the storage used does not provide mvcc_capability, this is why c._mvcc_storage is False. A stock FileStorage depends on the connection to provide MVCC (From: https://github.com/zopefoundation/ZODB/blob/master/src/ZODB/interfaces.py#L1022)

Since Relstorage supports MVCC at storage level, this code finally get executed and fails with multiple, nested databases. Which you already have with a stock Zope Site, because of the tempstorage.

do3cc:
@mgedmin I got a working test that triggers an error without my fix
@tseaver I've updated the changelog.

I've added comments to my test and my change in github about my understanding of what I am doing. I hope I got it all right.

tseaver:
@do3cc Thanks for your efforts!

PR #51, opened 2016-04-03
patch time.time in historical_connections to avoid windows failures.

jamadden:
Untested locally, copy-n-pasted from DemoStorage.test. Let's see what Travis says.

### DIFF
--- a/src/ZODB/historical_connections.txt
+++ b/src/ZODB/historical_connections.txt
@@ -11,10 +11,16 @@ Historical Connections
     ...     return _now
     >>> import time
     >>> real_time_time = time.time
+    >>> real_time_sleep = time.sleep
+    >>> def faux_time_sleep(amt):
+    ...    global _now
+    ...    _now += amt
     >>> if isinstance(time,type):
     ...    time.time = staticmethod(faux_time_time) # Jython
+    ...    time.sleep = faux_time_sleep
     ... else:
     ...     time.time = faux_time_time
+    ...     time.sleep = faux_time_sleep
     >>> def utcnow():
     ...     return datetime.datetime.fromtimestamp(_now)
 
@@ -312,6 +318,7 @@ memory usage would also be brief and unlikely to overlap.
 .. restore time
 
     >>> time.time = real_time_time
+    >>> time.sleep = real_time_sleep
 
 .. ......... ..
 .. Footnotes ..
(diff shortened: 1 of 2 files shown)


jamadden:
Well now they're all failing consistently with the original error (got `[False,False,False,True]` instead of `[False,False,True,True]`).

I think the problem is that the code is assuming that the two actions in line 233 in this block happens simultaneously:

    >>> db2.setHistoricalTimeout(.01)
    >>> time.sleep(.1)
    >>> conns2[2].close(); conns2[3].close() # 233

If the time increments more than .01 between the two `close` calls, then `conns2[2]` gets destroyed when `conns2[3]` gets closed.

Let me try reducing the interval at which `time.time` increments, I guess.

jamadden:
Adjusting the timing exposed a real bug in `availableGC` for historical pools, which only seems to be exposed on Python >= 3.4. Fixed in b68ac8e.

tseaver:
> Adjusting the timing exposed a real bug in availableGC for historical pools, which only seems to be exposed on Python >= 3.4. Fixed in b68ac8e.

W00t!

jamadden:
Should I squash the commits? GitHub can now do that on merge automatically, too.

tseaver:
Squashing should be fine, I think.  The feature is enabled on this repository:  you get a choice between the traditional merge commit and a squash merge on the "confirm merge" button.

jamadden:
Does that count as a "LGTM" (i.e., should I go ahead and merge)?

tseaver:
LGTM :)

jimfulton:
This change broke the test on Mac OS X.  Investigating. :(

src/ZODB/historical_connections.txt
```diff
@@ -2,6 +2,28 @@
...
+    ...    global _now
+    ...    _now += amt
+    >>> if isinstance(time,type):
+    ...    time.time = staticmethod(faux_time_time) # Jython
+    ...    time.sleep = faux_time_sleep
+    ... else:
+    ...     time.time = faux_time_time
+    ...     time.sleep = faux_time_sleep
+    >>> def utcnow():
+    ...     return datetime.datetime.fromtimestamp(_now)
```
jimfulton: datetime.fromtimestamp is defined to return localtime.

jimfulton:
The failure should occur on any machine who's system time isn't UTC.

It looks like a fix is straightforward. I'll make a new pr.

PR #54, opened 2016-04-25
Cleanup dependencies, remove ZEO leftovers

jmuchemb:
About the README, dependencies are listed twice. I changed it to list them only once and I updated this list. Given that it's error-prone, I'm not sure it's useful to be so verbose about prerequisites/installation (but that's another topic).


tseaver:
LGTM.  I verified that the removed ZEO docs are [present in the ZEO repo](https://github.com/zopefoundation/ZEO/tree/master/doc).

Issue #50, opened 2016-04-01
``load(oid)`` must die!


jimfulton:
**Update**: The discussion below primarily pertains to non-RelStorage storages.  We ended up moving to a model where ZODB's MVCC implementation moves to an adapter applied to storages other than RelStorage. The discussion below pertains (now) to how that adapter interacts with storages.  ZODB connections will still use `load` to load objects from `IMVCCStorage` instances, which will either be RelStorage instances or adapted instances.

<hr>

Currently, we use `load(oid)` to load current data for an object
unless we've recieved an invalidation for it (the object).  This leads to
complex code that:
- checkes to see if an object has been invalidated and if not:
- loads the object using `load`, and
- if while loading, the object was invalidated, it then reloads using
  `loadBefore`.

The `IStorage.load` method is problematic because it's required to
load current data.  This leads to delicate ordering of `load`
results and invalidations, which also provide information about the
current version of an object.  In FileStorage, we also have to get a
read lock on the file before we load current data, to prevent writes.

In reviewing NEO, I realized that there's no reason for `load` to
exist, and in fact, NEO doesn't really implement it.

First of all, it makes no sense to use `load` if the transaction
time (`Connection._txn_time`) is known.  We should just use
`loadBefore` with the transaction time.

Second, one can choose a suitable transaction time at the beginning of
a transaction.  A suitable transaction time is just past (because we
use `loadBefore`) the database transaction time.  We don't want to
read any data that was written later, regardless of what we've gotten
invalidations for.  We can use `IStorage.lastTransaction()` to
determine the last transaction id(/time). For storages with remote
data, like ZEO or NEO, we might choose to request the last transaction
id from the remote server to make sure we've processed any in-flight
transactions.  This is what NEO does.  (More about this in a separate
issue.)

By determining the transaction time at the start of a transaction, we
can:
- Always call `loadBefore` which can be much more efficiently
  implemented than `load()`.
- Greatly simplify `Connection._setstate`.
- Greatly simplify logic of ZEO, and perhaps NEO or RelStorage.

So, in conclusion:
- Reimplement transaction handling in Connection to set `_txn_time`
  at the beginning of a transaction using
  `p64(u64(storage.lastTransaction()) + 1)`.
- Simplify `_setstate` by always calling `loadBefore` using `_txn_time`.
  
  Update `get(oid)` in a similar fashion.  (This is probably not as big
  an issue in this case, as `get(oid)` doesn't load object state,
[...]

PR #56, opened 2016-05-04
Simplify MVCC by determining transaction start time using lastTransac…

jimfulton:
…tion.

This implements: https://github.com/zopefoundation/ZODB/issues/50

Rather than watching invalidations, simply use 1 + the storages
lastTransaction, which is equivalent to but much simpler than waiting
for the first invalidation after a transaction starts.

More importantly, it means we can always use loadBefore and get away
from load.  We no longer have to worry about ordering of invalidations
and load() results.

Much thanks to NEO for pointing the way toward this simplification!

Implementing this initially caused a deadlock, because DB.open()
called Connection.open() while holding a database lock and
Connection.open() now calls IStotage.lastTransaction(), which acquires a
storage lock. (It's not clear that lastTransaction() really needs a
storage lock.)  Meanwhile, IStotage.tpc_finish() calls a DB function
that requires the DB lock while holding the storage lock.  Fixing this
required moving the call to Connection.open() outside the region where
the DB lock was held.

To debug the problem above, I greatly improved lock-debugging
support. Now all of the ZODB code imports Lock, RLock and Condition
from ZODB.utils. If the DEBUG_LOCKING is set to a non-empty value,
then these are wrapped in such a way that debugging information is
printed as they are used. This made spotting the nature of the
deadlock easier.

Of course, a change this basic broke lots of tests. Most of the
breakage arises from the fact that connections now call
lastTransaction on storages at transaction boundaries.  Lots of tests
didn't clean up databases and connections properly.  I fixed many
tests, but ultimately gave up and added some extra cleanup code that
violated transaction-manager underware (and the underware's privates)
to clear transaction synchonizers in test setup and tear-down.  I plan
to add a transaction manager API for this and to use it in a
subsequent PR.

This tests makes database and connection hygiene a bit more important,
especially for tests, because a connection will continue to interact
with storages if it isn't properly closed, which can lead to errors if
the storage is closed.  I chose not to swallow these errors in
Connection, choosing rather to clean up tests.

The thread debugging and test changes make this PR larger than I would
have liked. Apologies in advance to the reviewers.


jimfulton:
The fix in the last commit was lame. I'm working on a better one.

jimfulton:
Still failures on travis. whimper.

jimfulton:
I'm going to close this PR.  More thought is needed wrt the semantics of loadBefore.  As pointed out in a comment in ClientStorage's loadBefore implementation, if loadBefore is used to get current data, then it needs to get the same care wrt invalidation ordering that load gets now.

jimfulton:
OK, more thought was given.

I think the ZEO issue I raised can be addressed in the asyncio work.

Relstorage maybe needs more thought.  It appears that despite it's popularity, that relstorage is orphaned. :( I think the changes here can make relstorage a lot simpler.

It would be great if someone(s) would review this and give a thumbs up/down without merging.

I plan to make a release of the current ZEO that requires ZODB < 5.0. Later, (if/)when this is merged, it will be ZODB 5.x.

jimfulton:
Gaaaa. This branch causes a lot of read conflicts in RelStorage tests. I'm going to need to sort that out. :(

jimfulton:
OK, I made a PR against RelStorage, https://github.com/zodb/relstorage/pull/31, to fix this.  RelStorage's lastTransaction was broken and it's monkey patch of ZODB caused problems.  RelStorage could be simplified a lot, especially with the changes in this branch. Maybe someday I'll fork it.

Please someone review this. :)

jimfulton:
I really need a reviewer for this...

src/ZODB/interfaces.py
```diff
@@ -577,6 +577,20 @@ def __len__():
...
+        future.  It is no longer used by ZODB, although it may still
+        be used in some tests or scripts.  Previously, there was a
+        requirement that load results be properly ordered with
+        invalidations so that at any point in time, clients have a
+        consistent view of what version of an object is current.  This
+        restriction has been relaxed and some storages will be
+        simplified as a result of the removal of this requirement.
+
+        An alternative to calling load is calling loadBefore passing
+        ZODB,utils.maxtid::
```
vpelletier: Typo: "ZODB," Comma instead of dot.

vpelletier:
I read changes except tests, and only found a ultra-minor typo in load's docstring. The rest looks fine to me, but I'm not yet up to speed on ZODB.

jimfulton:
Thank you very much @vpelletier! I'm surprised by "I'm not yet up to speed on ZODB". Aren't you one of the NEO authors?

BTW, one of my goals over the next few months is to simplify ZODB and improve documentation so coming up to speed will be easier.

vpelletier:
I mean I have barely touched neo these last... 3 years ? I forgot a lot of edge-cases I'm sure - although I think I still have a clear view of the overall design.

src/ZODB/interfaces.py
```diff
@@ -577,6 +577,20 @@ def __len__():
...
+        invalidations so that at any point in time, clients have a
+        consistent view of what version of an object is current.  This
+        restriction has been relaxed and some storages will be
+        simplified as a result of the removal of this requirement.
+
+        An alternative to calling load is calling loadBefore passing
+        ZODB.utils.maxtid::
+
+            store.loadBefore(oid, ZODB.utils.maxtid)
+
```
jensens: If this method is deprecated I'd love to see a deprecation warning. We (as the Plone community) documented common strategies and rationales at http://docs.plone.org/develop/styleguide/deprecation.html

jimfulton: Normally, I'd agree, but this isn't an API called by anything but ZODB itself. (Or more precisely it's tests.)  It doesn't make sense to generate warnings that users can't do anything about.

This  comment is aimed at storage implementers.

I could have changed all of the call sites in ZODB and ZEO call loadBefore instead.  That would probably be a good idea, but in a separate PR.  This one is big enough as it is. :)

jimfulton:
It looks like this has has one, maybe 2 reviews. Are there any objections to merging?

I'd like to do a couple of follow-on PRs:

- Convert the locking try/finallys to with statements

- Remove load calls from the ZODB code base, initially by just changing them to use loadBefore with maxtid.  In the unlikely event that there are unknown strorages in the wild that reuse storage tests from ZODB, then this should flush them out.  I'd also change the existing load implementations in the ZODB code base to use loadBefore.

Both of these will touch lots of lines of code, but will be pretty mechanical, so hopefully easy to review.

jimfulton:
I should note that there are branches of ZEO and RelStorage that work with (tests pass with) this branch.

src/ZODB/DB.py
```diff
@@ -628,7 +634,11 @@ def close(self):
 
         @self._connectionMap
         def _(c):
-            c.transaction_manager.abort()
+            if c.opened:
+                c.transaction_manager.abort()
+                # Note that this will modify out pool, but this is safe, because
```
jamadden: "modify our pool"?

jamadden:
Other than the tiny, minor typo comment, I have no objections to merging. I can't say for sure that I completely understand all the implications, but I am greatly reassured by how generally minor the test changes needed to be.

jmuchemb:
I have. Sorry, I planned to answer today, and again I got interrupted. Certainly this evening.

jimfulton:
@jmuchemb you have an objection to merging?

jamadden:
Perhaps slightly unrelated: would the ZODB5 release cycle (as a major release) be a good time to remove some long-deprecated things? I'm thinking primarily of the various "version" arguments, but there may be other things. It looks like most uses of `version` (with maybe the restore() method as an exception) are probably internal. Or is that too disruptive?

jimfulton:
I'm thinking a) I don't want to make this release any bigger than it is, and b) I think it would be disruptive.

I think it's been deprecated long enough that we could get rid of it later without a major version bump.

jmuchemb:
On the code, I disagree with commit 2729f3a928ea1bf0cc73cce56bd6bc77c032db30: `_txn_time` can't be set after retrieving oids to invalidate. Between these 2 points, there's a race condition when we receive invalidations. As soon as it is clarified, I'll run the NEO test suite with this branch.

On the interface, I'd like to go further. In the case of ZEO/NEO, sometimes we want to ping the server, and sometimes not. Currently, it's unclear whether lastTransaction() should/must ping.
- As mentioned in [a NEO monkey-patch](https://lab.nexedi.com/nexedi/neoppod/blob/e57ef6ccdf434ca2579ae4b12aa79c83ac0b88a1/neo/client/__init__.py#L66), we don't want a useless round-trip after a commit.
- On the other side, we have many applications where we want an up-to-date view of the DB at the moment the transaction begins. For ERP5, it's [required](https://lab.nexedi.com/nexedi/erp5/blob/f537d4e933a6f6f750e4422e2d09259166fd7046/product/CMFActivity/Activity/SQLBase.py#L497). For a server that only processes HTTP requests for humans, it may be useless.

So for example, we could introduce a new `IStorage` method, that would return the last tid, but with a boolean to tell whether it's for a new transaction or not.

jimfulton:
Good catch wrt _txn_time.  I caught that at the storage later but missed it here. I'll fix that.

jimfulton:
WRT pinging the server, by though was that a) this would be done in sync, and b) It would be controlled via storage configuration.

I'm not sure I understand your monkey patch, but it seems to imply that you're avoiding the server ping in afterCompletion, however, newTransaction is almost never called. (It's basically useless.)  The only reliable indication we have of transaction boundaries is afterCompletion, so that's sadly, the only place I can see to properly start the next transaction.

It might be useful later to add an option to TransactionManager.commit to signal that we're not going to start a new transaction and thus avoid some work.

Or maybe I'm misunderstanding.

jmuchemb:
Hmm... maybe I can change lastTransaction in NEO to not do a request to the master. After all, the client part maintain a copy of the last tid that is updated on each invalidation. And do the actual ping in sync. I haven't looked at the ZEO code but it would have to do the same.

> newTransaction is almost never called

It is. By `TransactionManager.begin()` since the connection is registered. And this method is the best way to control when to refresh the view of the DB.

So that's the interface of `sync` that would need to be reviewed, to take a boolean (new or after transaction), maybe rename it. In any case, it's currently missing from interfaces.py for non-MVCC storages.

jmuchemb:
About the deprecation of `IStorage.load()`, it is actually still used by ZODB, in `DB.__init__`

jmuchemb:
Instead of a boolean parameter to sync(), another option is to not call it at all from afterCompletion. tcp_finish already syncs the db, and the time elapsed until afterCompletion is negligible.

jimfulton:
Yes, ``newTransaction`` is called from ``begin``, but ``begin`` is rarely called. 

``newTransaction`` is also called by ``Connection.open``, which is enough for most web apps, for which open connections only execute a single transaction.

We could also update ``sync`` to call begin.

Some other candidates include:

- Anything that explicitly executes transactions, like context managers and decorators.

- Logic that retries transactions after transient errors, like repose retry.  It would be nice if the transaction package did more of this and tools like repos.retry just called that.

Note that tpc_finish isn't called unless there have been writes in a transaction, so we can't assume that a sync has just been done in afterCompletion. It's a shame that the completed transaction isn't passed in afterCompletion.

In general I like the drift of this discussion:

- Be more explicit about transaction boundaries and prefer begin over abort when you plan to start a new transaction.

- Do less in afterCompletion.

WRT the storage API:

- lastTransaction shouldn't mutate. It should be a pure getter.

- sync should do what it's name implies and sync with remote servers.

jimfulton:
Gaaa, wtf with PyPy and that last change....

jimfulton:
Relating to the sync/lastTransaction discussion, I've been talking to Shane and Jason about RelStorage and have a better understanding of what he was doing with IMVCC storage.  This has led me to believe that the best way to unify RelStorage with everything else is to split the ZODB MVCC logic into a storage later/adapter that's applied to anything that doesn't implement IMVCStorage (or something similar) itself.  I plan to write a note to the ZODB list about this today.  This doesn't really change anything we've talked about here, as load would still disappear from the storage API.

jimfulton:
Oh, the transaction *is* passed to afterCompletion.  So if the transaction remembered the last transaction it finished, it could avoid syncing in afterCompletion. (Why did I think it wasn't?) So maybe that's enough to avoid excess syncs.  We'd still do excess syncs after aborts, but I think it would be better not to sync in afterCompletion anyway.

jimfulton:
It looks like well-behaved transaction executors:

- The context managers in the transaction package

- zc.zodbwsgi :)

Call begin.

DB's context manager can be fixed. :)

The decorator I threatened here: https://groups.google.com/forum/#!topic/python-transaction/3eUzBzX2sRU can be written to do the right thing.

So given that, I'm inclined to:

- separate newTransaction from afterCompletion

- stop calling newTransaction in Connection.open, but *do* everything that newTransaction would do minus the sync.

jmuchemb:
Yes, I was also about to reply that for example, the context manager from transaction calls begin.

That depends on the way you use transaction.

For a sequence of transactions, without caring about what's done in parallel by any other process, like in unit tests, it's handy view what has been committed so far without calling begin(). But for web apps, it's not enough.

> stop calling newTransaction in Connection.open, but do everything that newTransaction would do minus the sync.

Maybe I miss something but sync() is useful when opening a Connection. See [this ERP5 monkey-patch](https://lab.nexedi.com/nexedi/erp5/blob/a0b62b51cda4a5ca392206b6d68daf7e03ba2d4d/product/ERP5Type/patches/ZODBConnection.py#L43), which we enables for example when doing mechanize test on a cluster of zopes behind haproxy.
Or maybe that's a problem in ZPublisher.

jimfulton:
I'm not sure I understand, but ZPublisher calls begin, which would cause newTransaction to be called, which would cause sync to be called.

If we sync on open, then in a typical web application, we'd end up syncing twice for each request, on open, and on begin. I thought you'd want to avoid that.

If not syncing on open causes problems for tests, maybe it would be best to fix the tests. Perhaps there are ways to tweak the APIs to make this easier. For example, the ``connection`` DB and storage convenience methods could call sync on the new connection. Many tests use these.  Or maybe ``DB.open`` could grow a sync option.  I'd prefer to keep this explicit so behavior in tests mirrors run-time behavior.

BTW, I'd change Connection.sync to use begin rather than abort.

jmuchemb:
I am currently reproducing a case where:
1. `transactions_manager.begin()` (ZPublisher/Publish.py:113) does nothing because there's no connection registered
2. just after in `request.traverse(...)` (line 115), at some point `__bobo_traverse__` calls `db.open()` which in turns calls `result.open(transaction_manager)`
3. in ZODB 3.10, `Connection.open` registers (`transaction_manager.registerSynch(self)`) but does not catch up by calling `newTransaction` (or any else that makes sure we have an up-to-date view of the DB), which can be an issue for some apps.

jimfulton:
Whimper. So the connection is being opened in traversal.  It's being opened in the middle of a transaction.  I suppose this is a valid use case.  If we sync on open, we're going to be penalizing apps that call begin after open. I wonder what Pyramid does.

jimfulton:
@mcdonc does Pyramid call TransactionManager.begin after opening a connection? Or before?

jamadden:
Pyramid doesn't have a dependency on transaction itself, it never touches a transaction or TransactionManager. Scaffolds and/or middleware/tweens have to do that.

jimfulton:
Sure, but what to the middleware or tweens do? I assume there's a standard/common configuration.

jamadden:
pyramid_tm  is one tween that does handle transactions and retry. But it doesn't know anything about ZODB. It calls begin (before traversal) but that may be before or after a Connection is open, depending on how it's configured. 

The ZODB scaffold that ships with pyramid stacks pyramid_tm below pyramid_zodbconn, so the transaction is begun *after* the connection is open.

I'll have to check what my groups stack does because we don't use either of those projects, we have a custom tween for both purposes.

jmuchemb:
Isn't it the purpose of a pool to reuse connection and saves the work done by open() ? In this case, syncs done by open() would be negligible.

jamadden:
Our tween stack does the same thing, the transaction is begun after the connection is opened.

jimfulton:
I suspect that opening a connection in the middle of a transaction is pathological. If that happens, then different data managers in the transaction have different notions of transaction start time, hurting consistency. (Of course, consistency in distributed transactions is a bit tenuous.)

jimfulton:
@jmuchemb I'm not sure I understand your last comment.  The ZODB connection pool reuses work done by __init__, and more importantly, the object cache.  The work I'd like to save here, largely in response to your comments earlier :), is the extra round trip to the storage server that might be done by sync.

jimfulton:
Perhaps the ZODB configuration should grow a sync_on_open that does a storage sync when a connection is opened that would be set in Zope 2 applications.

tseaver:
FTR, here is the `pyramid_tm` tween [calling `begin()`](https://github.com/Pylons/pyramid_tm/blob/5b7e08414ccc9f4063239917624d06986643c7ee/pyramid_tm/__init__.py#L70):  as a tween, this call happens before Pyramid ever starts traversal.

jmuchemb:
@jimfulton Ok, I didn't know that `open()` was called all the time when processing requests with ZPublisher. For apps that begin the transaction after the db is open, it is indeed a problem to sync in `open()`

jmuchemb:
What about testing from Connection.open whether the transaction has begun or not ?
Or rather change `TransactionManager` as follows:
```python
    def registerSynch(self, synch):
        """ See ITransactionManager.
        """
        self._synchs.add(synch)
        if self._txn is not None:
            synch.newTransaction(self)
```

jimfulton:
That's a great idea. It makes the transaction behavior more complete.  I will make it so.

jimfulton:
I made it so: https://github.com/zopefoundation/transaction/pull/16

vpelletier:
I did not catch up with all posts yet, but closely related to the `load()` call in `DB.__init__`, I have noticed the following:
When OID 0 does not exist, it is created with `serial=None`, but when creating an object (in `Connection._store_objects`) it gets created with `serial=z64`.

Please pick one, and use in both places.

About which to pick, I vote for `z64`, seeing OID 0 as the exception currently and all other objects as the rule.

jimfulton:
@vpelletier I mostly agree, but I think that should be a separate PR.  Note that for new objects, the serial argument is ignored. I'd rather use None as the serial for new objects, as, in my mind, it makes it more explicit that we're dealing with a new object.

jimfulton:
OK, in the last commit I arranged that storage ``sync`` is only called if ``begin`` is called or if a data manager registers with a transaction manager while a transaction is active.  The later case addresses Zope 2's somewhat odd behavior.

I'd like to get this PR wrapped up. I have a couple of cleanup branches backed up behind it.

I've also decided to take a stab at an IMVCCStorage adapter so that ZODB5 can support RelStorage.

Please, someone merge this PR or tell me I can. :)

setup.py
```diff
@@ -23,25 +23,9 @@
...
-    sys.exit(0)
-
-
-# The (non-obvious!) choices for the Trove Development Status line:
-# Development Status :: 5 - Production/Stable
-# Development Status :: 4 - Beta
-# Development Status :: 3 - Alpha
-
 classifiers = """\
-Development Status :: 4 - Beta
```
tseaver: Shouldn't we label it as `Production/Stable`?

src/ZODB/Connection.py
```diff
@@ -866,109 +877,42 @@ def setstate(self, obj):
...
-
-        # MVCC Blob support
-        if isinstance(obj, Blob):
-            obj._p_blob_uncommitted = None
-            obj._p_blob_committed = self._storage.loadBlob(obj._p_oid, start)
-
-        return True
+        except ConflictError:
+            raise
+        except:
```
tseaver: Shouldn't this be logging something about the actual exception?

src/ZODB/Connection.py
```diff
@@ -1083,6 +1042,7 @@ def _resetCache(self):
             self._reader._cache = cache
 
     def _release_resources(self):
+        assert not self.opened
```
tseaver: I guess this is a place where our styles clash:  I don't like `assert` in non-test code.

src/ZODB/DB.py
```diff
@@ -753,18 +763,18 @@ def open(self, transaction_manager=None, at=None, before=None):
             assert result is not None
 
             # open the connection.
-            result.open(transaction_manager)
```
tseaver: Comment should go, too.

src/ZODB/tests/testmvcc.py
```diff
@@ -33,9 +33,17 @@
 test that the implementation is correct).  The tests use a
 MinimalMemoryStorage that implements MVCC support, but not much else.
 
+***IMPORTANT***: The MVCC approach has changed since these tests were
+originally written. The new approach is much simpler because we no
+longer call load to get the current state of an object. We call
+loadBefore instead, having gotten a transaction time at the start of a
+transaction.  As a result, the rhythm of the tests is a little odd,
+because the probe a complex dance that doesn't exist any more.
```
tseaver: Word(s) missing?

setup.py
jimfulton: I think this is just a waste of time. Will this be updated if we make an alpha or beta release? I doubt  it.  The development status is expressed in the version number.  This is a DRY violation.

I'm a strong -1 on fooling with this meta data.

tseaver: OK.  FWIW, I read this as a signal about the state of the *project*, rather than a given release.

src/ZODB/Connection.py
jimfulton: Logger._exception does that. I assume you're referring to the line below your comment.

Note that this is code I didn't actually touch. The diff is misleading. I folded multiple functions together.
See lines 868-875 in the old code.

jimfulton: As a rule, I try not to make changes unrelated to what I'm doing to try to limit the sizes of PRs.

src/ZODB/Connection.py
jimfulton: This code has lots of asserts.  I'm ambivalent about them myself, but again, they're used a lot in ZODB, so I played along. <shrug>

I think they're a terrible idea in test code, unless you're using py.test.

src/ZODB/tests/testmvcc.py
jimfulton: I reworded a bit.

src/ZODB/DB.py
jimfulton: Yup, fixed.

setup.py
jimfulton: I just tried to find a definition of this field and couldn't find one.  Maybe I should have searched harder. <shrug>
Given the ambiguity. all the more reason to avoid it.  I certainly thought it related to releases, but I was probably wrong.

FWIW I typically interpret 0.x releases as less than mature.  People (myself included) often are too lazy to make 1.x releases when they deem something mature.  These people aren't going to make releases to update the trove status.  BTW the trove status: "Development Status :: 7 - Inactive" makes me laugh. :)

jmuchemb:
Oops, when suggesting to change `TransactionManager.registerSynch`, I didn't realize that we'd still have to keep a call to `newTransaction(sync=False)` inside `open()`. Which could have been avoided with an API to test whether the transaction has begun or not. Maybe it's cheap enough so I can live with that.

In the comment, there's a typo: "pathalogical"

I could run NEO unit tests with this branch. Obviously, I had to do a few changes (not yet committed). All tests pass without hacking.

jimfulton:
Yeah, the ``newTransaction`` call is needed for tests :) and more importantly clients that don't call ``begin`` and don't open a transaction while a transaction is active.

The call to newTransaction is cheap and if it does any work, the following call won't have to.

Cool WRT NEO tests!

jimfulton:
OK, so reviewers seemed happy this morning.  I merged https://github.com/zopefoundation/ZODB/pull/66  and resolved conflicts.  This should increase happiness.

This is known to work with RelStorage master, which required minor changes, essentially bug fixes.

If there are no objections, I'll merge this tomorrow late afternoon.

I won't release this until after RelStorage 1.7 is released. (ZEO 4.2 was released today and works with it.)

jamadden:
> I won't release this until after RelStorage 1.7 is released.

I was hoping to accomplish at least a basic Python 3 port for RelStorage 1.7 (among a few other things), but that's probably at least a couple days of work. That'd be kind of a consolation gift for technically dropping support for <= Python 2.6 (due to the updated ZODB dependency needed for PyPy support).

However if it's important to get a ZODB5 release out sooner rather than later, that *could* be put off for a  1.8 release. Other than a few documentation updates, RelStorage is in a releasable state and could go out as early as mid-morning tomorrow, if necessary.

jimfulton:
I don't have to release ZODB5 until I'm ready to release ZEO5.  So there isn't a rush to release.  I don't want to hold off the release too long mainly to avoid bitrot.

src/ZODB/mvccadapter.py
```diff
@@ -0,0 +1,256 @@
...
+    def __getattr__(self, name):
+        if name in self._copy_methods:
+            if hasattr(self._storage, name):
+                m = getattr(self._storage, name)
+                setattr(self, name, m)
+                return m
+
+        raise AttributeError(name)
+
+    def __len__():
```
jmuchemb: missing self

jimfulton: Oops. Thanks! Fixed.

src/ZODB/Connection.py
```diff
@@ -490,62 +440,6 @@ def _tpc_cleanup(self):
...
-            # this is to, in fact, make classes ghostifiable.  Then
-            # we'd have to reimplement attribute lookup to check the
-            # class state and, if necessary, activate the class.  It's
-            # much worse than that though, because we'd also need to
-            # deal with slots.  When a class is ghostified, we'd need
-            # to replace all of the slot operations with versions that
-            # reloaded the object when called. It's hard to say which
-            # is better or worse.  For now, it seems the risk of
-            # using a class while objects are being invalidated seems
-            # small enough to be acceptable.
```
jmuchemb: There were interesting things in this comment. Shouldn't it be kept somewhere ?

PR #57, opened 2016-05-04
On Python 2 and Mac OS X, checkFlushAfterTruncate was failing.

jimfulton:
Changed to always empty the file pool.  I doubt that just flushing on
Python 2 was much of an optimization, and ... Python 2. :)


jmuchemb:
You can remove the now unused import of PY3.

jmuchemb:
I also prefer to keep the comment.

jimfulton:
The comment wouldn't make sense without the version check. I'll adjust it.

tseaver:
LGTM.  I believe this change will fix the broken Python 2.x Windows tests as well.

PR #61, opened 2016-05-17
Enforce minimium required Python versions: 2.7, 3.3.

### LINKED ISSUE #59: setup.py classifiers not inline with sys.version_info checking

tseaver:
Closes #59.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,22 +23,7 @@ interface, rich transaction support, and undo.
 version = "4.3.0.dev0"
 
 import os
-import sys
 from setuptools import setup, find_packages
-
-if sys.version_info < (2, 6):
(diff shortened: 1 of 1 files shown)


tseaver:
The other option would be to remove the `sys.version_info` checks altogether.

tseaver:
`pypy3` is an anomaly:  its `sys.version_info` reports `3.2.x`, but it supports the unicode literal (a 3.3.x feature), so the tests still pass.

I'd rather not make the `sys.version_info` check more complicated (e.g., testing `platform`).  @jimfulton WDYT about just removing the checks?

tseaver:
I've reworked this branch to drop the `sys.version_info` checks.

PR #62, opened 2016-05-30
Fixed a test that depended on some off behavior of tpc_finish in some…

jimfulton:
… storages.

The ZEO asyncio work depends on this fix.


tseaver:
LGTM

PR #66, opened 2016-06-14
Abstract ZODB's MVCC implementation into a storage adapter.

jimfulton:
That's applied to storages other than RelStorage.

See: https://groups.google.com/forum/#!topic/zodb/QJYmvF0eUUM

### DIFF
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -49,6 +49,8 @@ from ZODB.utils import p64, u64, z64, oid_repr, positive_id
 from ZODB import utils
 import six
 
+from .mvccadapter import HistoricalStorageAdapter
+
 global_reset_counter = 0
 
 noop = lambda : None
@@ -101,11 +103,10 @@ class Connection(ExportImport, object):
         # Multi-database support
         self.connections = {self._db.database_name: self}
 
-        storage = db._mvcc_storage
         if before:
-            storage = storage.before_instance(before)
+            storage = HistoricalStorageAdapter(db.storage, before)
         else:
-            storage = storage.new_instance()
+            storage = db._mvcc_storage.new_instance()
 
         self._normal_storage = self._storage = storage
         self.new_oid = db.new_oid
@@ -308,7 +309,7 @@ class Connection(ExportImport, object):
         """Returns True if this connection is read only."""
         if self.opened is None:
             raise ConnectionStateError("The database connection is closed")
-        return self.before is not None or self._storage.isReadOnly()
+        return self._storage.isReadOnly()
 
     @property
     def root(self):
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -103,10 +103,9 @@ class Connection(ExportImport, object):
         # Multi-database support
         self.connections = {self._db.database_name: self}
 
+        storage = db._mvcc_storage.new_instance()
         if before:
-            storage = HistoricalStorageAdapter(db.storage, before)
-        else:
-            storage = db._mvcc_storage.new_instance()
+            storage = HistoricalStorageAdapter(storage, before)
 
         self._normal_storage = self._storage = storage
         self.new_oid = db.new_oid
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -103,9 +103,17 @@ class Connection(ExportImport, object):
         # Multi-database support
         self.connections = {self._db.database_name: self}
 
-        storage = db._mvcc_storage.new_instance()
+        storage = db._mvcc_storage
         if before:
-            storage = HistoricalStorageAdapter(storage, before)
+            try:
+                before_instance = storage.before_instance
+            except AttributeError:
+                def before_instance(before):
+                    return HistoricalStorageAdapter(
+                        storage.new_instance(), before)
+            storage = before_instance(before)
+        else:
+            storage = storage.new_instance()
 
         self._normal_storage = self._storage = storage
         self.new_oid = db.new_oid
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -27,7 +27,6 @@ from persistent import PickleCache
 from persistent.interfaces import IPersistentDataManager
 from ZODB.interfaces import IConnection
 from ZODB.interfaces import IBlobStorage
-from ZODB.interfaces import IMVCCStorage
 from ZODB.blob import Blob, rename_or_copy_blob, remove_committed_dir
 from transaction.interfaces import ISavepointDataManager
 from transaction.interfaces import IDataManagerSavepoint
@@ -102,13 +101,11 @@ class Connection(ExportImport, object):
         # Multi-database support
         self.connections = {self._db.database_name: self}
 
-        storage = db.storage
-        if IMVCCStorage.providedBy(storage):
-            # Use a connection-specific storage instance.
-            self._mvcc_storage = True
-            storage = storage.new_instance()
+        storage = db._mvcc_storage
+        if before:
+            storage = storage.before_instance(before)
         else:
-            self._mvcc_storage = False
+            storage = storage.new_instance()
 
         self._normal_storage = self._storage = storage
         self.new_oid = db.new_oid
@@ -138,17 +135,17 @@ class Connection(ExportImport, object):
         # ReadConflictError (just to be able to clean them up from the
         # cache on abort with the other modified objects). All objects
         # of this list are either in _cache or in _added.
-        self._registered_objects = []
+        self._registered_objects = [] # [object]
 
         # ids and serials of objects for which readCurrent was called
         # in a transaction.
-        self._readCurrent = {}
+        self._readCurrent = {} # {oid ->serial}
 
         # Dict of oid->obj added explicitly through add(). Used as a
         # preliminary cache until commit time when objects are all moved
         # to the real _cache. The objects are moved to _creating at
         # commit time.
-        self._added = {}
+        self._added = {} # {oid -> object}
 
         # During commit this is turned into a list, which receives
         # objects added as a side-effect of storing a modified object.
@@ -162,32 +159,11 @@ class Connection(ExportImport, object):
         # adding. Used during abort to remove created objects from the
         # _cache, and by persistent_id to check that a new object isn't
         # reachable from multiple databases.
-        self._creating = {}
+        self._creating = {} # {oid -> implicitly_added_flag}
 
         # List of oids of modified objects, which have to be invalidated
         # in the cache on abort and in other connections on finish.
-        self._modified = []
-
-        # _invalidated queues invalidate messages delivered from the DB
-        # _inv_lock prevents one thread from modifying the set while
-        # another is processing invalidations.  All the invalidations
-        # from a single transaction should be applied atomically, so
-        # the lock must be held when reading _invalidated.
-
-        # It sucks that we have to hold the lock to read _invalidated.
-        # Normally, _invalidated is written by calling dict.update, which
-        # will execute atomically by virtue of the GIL.  But some storage
-        # might generate oids where hash or compare invokes Python code.  In
-        # that case, the GIL can't save us.
-        # Note:  since that was written, it was officially declared that the
-        # type of an oid is str.  TODO:  remove the related now-unnecessary
-        # critical sections (if any -- this needs careful thought).
-
-        self._inv_lock = utils.Lock()
-        self._invalidated = set()
-
-        # Flag indicating whether the cache has been invalidated:
-        self._invalidatedCache = False
+        self._modified = [] # [oid]
 
         # We intend to prevent committing a transaction in which
         # ReadConflictError occurs.  _conflicts is the set of oids that
@@ -197,12 +173,6 @@ class Connection(ExportImport, object):
         # will raise ReadConflictError again (because the oid is in
         # _conflicts).
         self._conflicts = {}
-
-        # _txn_time stores the upper bound on transactions visible to
-        # this connection. That is, all object revisions must be
-        # written before _txn_time. If it is None, then the current
-        # revisions are acceptable.
-        self._txn_time = None
 
         # To support importFile(), implemented in the ExportImport base
         # class, we need to run _importDuringCommit() from our commit()
@@ -252,17 +222,7 @@ class Connection(ExportImport, object):
         if obj is not None:
             return obj
 
-        before = self.before
-        if before is None:
-            # Normal case
-            before = self._txn_time
-
-        data = self._storage.loadBefore(oid, before)
-        if data is None:
-            # see the comment in setstate
-            raise ReadConflictError()
-
-        p, _, _ = data
+        p, _ = self._storage.load(oid, '')
         obj = self._reader.getGhost(p)
 
         # Avoid infiniate loop if obj tries to load its state before
@@ -320,9 +280,6 @@ class Connection(ExportImport, object):
         if self.opened:
             self.transaction_manager.unregisterSynch(self)
 
-        if self._mvcc_storage:
-            self._storage.sync(force=False)
-
         if primary:
             for connection in self.connections.values():
                 if connection is not self:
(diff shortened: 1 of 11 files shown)


jimfulton:
Travis is green.

jimfulton:
Found a ZEO-specific undo bug running ZEO tests against this branch. Fixing...

jimfulton:
Travis is green again.

jimfulton:
Can someone review this, please?

src/ZODB/mvccadapter.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def __len__():
+        return len(self._storage)
+
+class MVCCAdapter(Base):
+
+    def __init__(self, storage):
+        Base.__init__(self, storage)
+        self._instances = set()
+        self._lock = threading.Lock()
```
jamadden: Should that be the new ZODB.utils lock you built?

src/ZODB/mvccadapter.py
```diff
@@ -0,0 +1,268 @@
...
+        'isReadOnly', 'loadSerial', 'new_oid', 'tpc_vote',
+        'checkCurrentSerialInTransaction', 'tpc_abort',
+        )
+
+    def __init__(self, base):
+        self._base = base
+        Base.__init__(self, base._storage)
+        self._lock = threading.Lock()
+        self._invalidations = set()
+        self._start = None # Transaction start time
```
jamadden: `start_tid`?

src/ZODB/mvccadapter.py
```diff
@@ -0,0 +1,268 @@
...
+                if instance is not committing_instance:
+                    instance._invalidate(oids)
+
+    references = serialize.referencesf
+    transform_record_data = untransform_record_data = lambda self, data: data
+
+    def pack(self, pack_time, referencesf):
+        return self._storage.pack(pack_time, referencesf)
+
+class MVCCAdapterInstance(Base):
```
jamadden: Should this provide the same interfaces as the base storage, as well as IMVCCStorage? True `Connection._storage` is in implementation detail, but code in the wild might be looking at it.

src/ZODB/mvccadapter.py
jimfulton: Yup. I'll change that.

src/ZODB/mvccadapter.py
```diff
@@ -0,0 +1,268 @@
...
+        self._storage = storage
+        for name in self._copy_methods:
+            if hasattr(self._storage, name):
+                setattr(self, name, getattr(storage, name))
+
+        if interfaces.IBlobStorage.providedBy(storage):
+            zope.interface.alsoProvides(self, interfaces.IBlobStorage)
+
+    def isReadOnly(self):
+        return self._storage.isReadOnly()
```
jamadden: Maybe a comment about why these aren't being copied? (Because subclasses override.) It took me a minute to see that.

src/ZODB/mvccadapter.py
jimfulton: Oh here we go. :)

The thing is, it isn't a tid as it doesn't identify a transaction.  It's really a transaction start time.  I tried to avoid the whole tid/time issue by just calling it _start.

src/ZODB/mvccadapter.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def supportsUndo(self):
+        try:
+            return self._storage.supportsUndo()
+        except AttributeError:
```
jamadden: In order to avoid muffling real AttributeErrors, I tend to spell these things thusly:
```
try:
    meth = self._storage.supportsUndo
except AttributeError:
   return False
else:
   return meth()
```

If there are more than a handful of such methods, I tend to create a factory so I can write:
```
   supportsUndo = _proxy_to_storage('supportsUndo')
```
But that's a matter of taste.

src/ZODB/mvccadapter.py
```diff
@@ -0,0 +1,268 @@
...
+    def load(self, oid, version=''):
+        assert self._start is not None
+        r = self._storage.loadBefore(oid, self._start)
+        if r is None:
+            raise POSException.ReadConflictError(repr(oid))
+        return r[:2]
+
+    def tpc_begin(self, transaction):
+        self._storage.tpc_begin(transaction)
+        self._modified = set()
```
jamadden: My linter complains about setting attributes outside of __init__.

To preserve the same "you get an error if you call `store` outside of `tpc_begin`" behaviour, you could have `_modified` be a class variable that's a tuple or None. Plus that gives a handy place to document it :)

src/ZODB/DB.py
```diff
@@ -401,12 +401,6 @@ def __init__(self, storage,
           - `xrefs` - Boolian flag indicating whether implicit cross-database
             references are allowed
         """
-        if isinstance(storage, six.string_types):
-            from ZODB import FileStorage
-            storage = ZODB.FileStorage.FileStorage(storage, **storage_args)
-        elif storage is None:
-            from ZODB import MappingStorage
-            storage = ZODB.MappingStorage.MappingStorage(**storage_args)
```
jamadden: Why the removal of the convenience args? I know they don't seem to be documented but I bet they're in use. They don't seem to hurt much.

src/ZODB/interfaces.py
```diff
@@ -818,6 +650,137 @@ def tpc_vote(transaction):
...
+
+    def registerDB(wrapper):
+        """Register a storage wrapper IStorageWrapper.
+
+        The passed object is a wrapper object that provides an upcall
+        interface to support composition.
+
+        Note that, for historical reasons, an implementation may
+        require a second argument, however, if required, the None will
+        be passed as the second argument.
```
jamadden: MVCCAdapter breaks this promise, so it should probably either change or the documentation be updated.

src/ZODB/mvccadapter.py
jimfulton: Well, the only way that code in the wild can look at it is by accessing the connection's underware.  When you do that, you have to expect surprises. :)

These interfaces are a bit of a mess, owing in part to Shane's decision have new_instance clone the existing thing.  IMO, this makes RelStorage's implementation harder to reason about.   <shrug>  Another source of complexity is that both DB exposes a "storage" and connection uses a storage, but what they expose and use are different.  Historically, this didn't matter, because there were the same thing.

If we were really going to work this out, I might have IMVCCStorage and IMVCCConnection with distinct APIs that didn't extend IStorage. I'm willing to do this now, but I'd rather not because I have higher priorities.

src/ZODB/DB.py
```diff
@@ -438,12 +444,10 @@ def __init__(self, storage,
                 DeprecationWarning, 2)
             storage.tpc_vote = lambda *args: None
 
-        if IMVCCStorage.providedBy(storage):
-            temp_storage = storage.new_instance()
-        else:
-            temp_storage = storage
+        temp_storage = self._mvcc_storage.new_instance()
```
jamadden: Somewhere you recently said something about how the way IMVCCStorage clones itself is weird. I agree :) an IMVCCStorageFactory could clean this up and make the distinctions and responsibilities clearer.

src/ZODB/mvccadapter.py
jamadden: I'm ok with delaying the separation of APIs for now.

I do know that a number of RelStorage tests (and up until today, zodbconvert) depend on being able to access known attributes of the `Connection._storage`. I'm 80% sure that my own company's project's do the same, and I would bet that a lot of other's do too.

How about at least a `__getattr__` implementation to proxy things through?

If you do that, you wouldn't really be lying about the interfaces (much) if you copied them. Which is about as simple as saying `interface.alsoProvides(interface.providedBy(base))`, IIRC.

src/ZODB/interfaces.py
```diff
@@ -818,6 +650,137 @@ def tpc_vote(transaction):
 
         """
 
+    def close():
```
jamadden: This is just reorganizing the IStorage methods, nothing substantial, right?

src/ZODB/interfaces.py
jimfulton: OK, I'll fix the interface.

src/ZODB/tests/testConnection.py
```diff
@@ -521,29 +529,18 @@ def doctest_invalidate(self):
         Transaction ids are 8-byte strings, just like oids; p64() will
         create one from an int.
 
-        >>> cn.invalidate(p64(1), {p1._p_oid: 1})
+        >>> mvcc_storage.invalidate(p64(1), {p1._p_oid: 1})
```
jamadden: It might be nice if the invalidate and invalidateCache methods stayed on Connection (even if not in the interface) for BWC. Or in a `DeprecatedConnectionMethodsMixin` to help keep Connection itself smaller. They could easily simply pass the args to the storage.

src/ZODB/DB.py
jimfulton: Yes, except it's more than just a factory. I'd like to clean this up. I'd rather not do it right now.

src/ZODB/mvccadapter.py
jimfulton: I don't understand what you're suggesting.

src/ZODB/mvccadapter.py
jimfulton: OK

src/ZODB/mvccadapter.py
jimfulton: I think that's overkill here, but I defer.

src/ZODB/mvccadapter.py
jimfulton: OK

src/ZODB/interfaces.py
jimfulton: No, mostly a turd from an aborted attempt to factor the interfaces. I'll put it back.

src/ZODB/Connection.py
```diff
@@ -559,10 +497,6 @@ def _commit(self, transaction):
             elif oid in self._added:
                 assert obj._p_serial == z64
             elif obj._p_changed:
-                if oid in self._invalidated:
```
jamadden: This and line 603 are just what we talked about: not pre-emptively raising a ConflictError on invalidations, instead letting the normal conflict resolution process take care of that (or not), right?

I can see why this looks like an optimization, but unless it somehow turns out to destroy some workloads I still agree that the simplification is worthwhile.

src/ZODB/mvccadapter.py
jamadden: MVCAdapterInstance should have:
```python
def __getattr__(self, name):
   return getattr(self._base._storage, name)
```
and it's `__init__` could contain:
```
    interface.alsoProvides(interface.providedBy(self._base._storage)
```

src/ZODB/tests/testConnection.py
jimfulton: These were used internally by ZODB. Nothing should depend on them externally. They never should have been in IConnection.  I'd rather just consider their inclusion in IConnection a bug that this PR fixes.

jamadden:
Beautiful, I love it! I left some (ok lots of) comments, I hope they're not too bikeshed-y.

src/ZODB/tests/testConnection.py
jamadden: I agree they didn't make sense on Connection. I was just afraid that some app, probably Zope2 or Plone, with my luck, would be relying on them. But if you're comfortable removing them so am I.

src/ZODB/Connection.py
jimfulton: Yup.  Plus it reduces commit work a little bit for the common case, so could be a net win. <shrug>

jamadden:
I understand from your PRs against RelStorage that you've run the RelStorage tests against this branch?

jamadden:
(I should mention that I don't use historical connections or undo, so I'm not as familiar with that code, so someone should double-check me, but it still LGTM.)

src/ZODB/mvccadapter.py
jimfulton: OK, I'd really rather not.

- This just serves clients that grab underware.  I'd rather not encourage that.

- A future interface rationalization is likely to promise less, so I'd rather not provide more now.

Do you know of code in the would that would need this?

jimfulton:
Yes, I ran relstorage tests against this branch. 

The RelStorage tests test undo. I had to make a small change to address that. (For undo, I use a special instance that doesn't get invalidations and doesn't need to be registered.)

The RelStorage tests don't test historical connections, so I assume RelStorage doesn't support them. If it does, than we should discuss.  My hope was that connections wouldn't have a reason to call loadBefore.

src/ZODB/mvccadapter.py
jamadden: > Do you know of code in the [wild?] that would need this?

There are maybe 20 RelStorage tests that rely on something like `conn._storage.needs_poll()` and so forth. Now, those won't directly be affected because relstorage already implements IMVCCStorage, but it seemed like if RS did it, possibly others did too. (That does highlight a difference though, in the handling of IMVCCStorages vs not: you lose access to custom attributes in the "not" case that you used to be able to get.) I can imagine test code directly trying to get to the dicts maintained by a MappingStorage, for instance.

In NextThought's code, I found at least one place where we access `conn._storage.load`. That may be a bug, and it wouldn't be affected by this because the adapter provides `load` (but not `loadBefore` AFAICS), but here too, it seemed like if we did it, probably others did too.

Of course, the plural of "anecdote" is not data, and personally I tend to be conservative about potentially breaking changes even in implementation details.

jamadden:
I would assume that RelStorage supports historical connections; it defaults to preserving all history in the RDMBS, writing a new row for every revision of an object. I don't know why it would do that if it didn't support loadBefore and thus historical connections.

But yeah, I can't find it being explicitly tested anywhere.

jimfulton:
Hm, WRT historical connections.  I'll see if I can figure out a way to find a test I can reuse and add an adapter or other hack.

src/ZODB/mvccadapter.py
jimfulton: Done.

src/ZODB/mvccadapter.py
jimfulton: ZEO tests used conn_storage to access client storages, to access non-standard attributes.  These were easy to fix.  I regret neither the original code nor having to fix it. :)  (I have a non-standard enlightened attitude toward tech  debt.)

Application code accessing ``connection._storage.load`` is almost certainly a bad idea.  Breaking such code is good for it in the long run.

I'm not going to go out of my way to disallow access to underware, but I don't want to pander to it either.

src/ZODB/mvccadapter.py
jimfulton: Addressed by changing the copy mechanism: https://github.com/zopefoundation/ZODB/pull/66/commits/c7e99e3de151fee77f6d4c7ac7debe2d2ea4c274

src/ZODB/mvccadapter.py
jimfulton: Addressed by changing the copy mechanism: https://github.com/zopefoundation/ZODB/pull/66/commits/c7e99e3de151fee77f6d4c7ac7debe2d2ea4c274

src/ZODB/mvccadapter.py
jamadden: Fair enough.

I just now got the pun behind 'underware'.

src/ZODB/Connection.py
```diff
@@ -101,11 +103,10 @@ def __init__(self, db, cache_size=400, before=None, cache_size_bytes=0):
         # Multi-database support
         self.connections = {self._db.database_name: self}
 
-        storage = db._mvcc_storage
         if before:
-            storage = storage.before_instance(before)
+            storage = HistoricalStorageAdapter(db.storage, before)
```
jamadden: I'm slightly concerned about the asymmetry between the historical case and the normal case. I think I've convinced myself that the MVCC stuff (and keeping track of the instances in use) doesn't matter for the before connections because they're ReadOnly. Would this work instead as a simplification?

```python
storage = db._mvcc_storage.new_instance() # always get new instance
if before:
   storage = HistoricalStorageAdapter(storage, before)
```

jimfulton: I'm slightly concerned too, but the test passes.

I thought of the lack of new_instance after I pushed this. Thinking about it, I suspect it would break with multiple threads.

Your change (which is what I used initially) doesn't work unless I add loadBefore to the MVCCAdapterInstance, which I was trying to avoid.  I'll go ahead and add it for now.

jimfulton: OK, I came up with something that seems pretty sane, and allows me to keep loadBefore out of MVCCAdapterInstance. :)

jamadden: Seems reasonable to me. 

The try/catch has to stay for BWC for now, but maybe one day IMVCCStorage should include `before_instance`, which RelStorage would implement pretty much exactly like the function you defined.

PR #68, opened 2016-06-16
Load calls load before

jimfulton:
This is a cleanup PR that changes the storage ``load`` implementations to call ``loadBefore``. Eventually, these load implementations will go away, but lots of tests use them, so ...

A follow on PR will change the tests in ZODB to not call load. :)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,7 +2,12 @@
  Change History
 ================
 
-4.3.0 (unreleased)
+4.3.1 (2016-06-06)
+==================
+
+- Fixed: FileStorage loadBefore didn't handle deleted/undone data correctly.
+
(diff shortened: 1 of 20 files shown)


tseaver:
Should `ZODB.utils.load_current` be private?

jimfulton:
I don't think so, but I don't have a strong opinion at this point. Do you think it should be private?

jimfulton:
@tseaver ?

tseaver:
Dunno, I was just seeing it newly added, and didn't know whether we wanted/expected outside callers to use it:  if so, it could use a docstring.

jimfulton:
Docstring added

jimfulton:
Travis is green :)

tseaver:
LGTM

PR #69, opened 2016-06-17
Allow serial to be returned as late as tpc_finish

jmuchemb:
Because of ZODB wrappers like ZEO, I can't touch FileStorage & DemoStorage without breaking compatibility. So what I suggest is to first develop the 2 API in the 4.x branch, with deprecation warnings. Once done, we merge into master branch and (I prefer) force to use the new API.


jmuchemb:
Any idea about the PyPy failures? On my machine, I have no issue with PyPy 5.3
If it's a PyPy bug, do we care about such an old version (2.4.0) ?

jimfulton:
IIRC, I saw these errors on an unrelated branch. They're intermittent.

We should probably try to figure out WTF at some point,but I don't think they have anything to do with this PR.

jmuchemb:
I found the issue without actually trying to find why PyPy failed. There was really a race condition in my PR,  and PyPy 5.3 also failed randomly (although not very often).

jmuchemb:
I didn't realized that there were no test checking the returned values of store/tpc_vote in the case of conflict resolution. More work is needed and I'll do that in separate commits.

jimfulton:
"More work is needed and I'll do that in separate commits." I take this to mean that you'll be doing more work on this PR. I'll hold off reviewing for now.

Thanks for this!

jmuchemb:
I have finished for the 4.x branch. The first 2 commits are the minimum.
As written earlier, I prefer to force the new API in ZODB 5, since maintaining compatibility is not helping anyone. For example, we can 't use NEO with any wrapper that uses the old API. So:
- the 2nd commit also changes DemoStorage.tpc_finish to forward the commit tid
- the 3rd one deprecates the old API

And in the 4th one, I take this opportunity to optimize the returned value of store/tpc_vote. Like for tpc_finish, this new API would be mandatory in ZODB 5.

If all commits are accepted, then I'll work on merging into ZODB 5:

1. resolve conflicts
2. update storages & tests to use the new API
3. drop old API

jimfulton:
Please let me know when this is ready for review.

jmuchemb:
It is.

jimfulton:
This LGTM. Thanks for this.

Let's review the transition plan.

We make a ZODB 4 release with these changes.

We then find, um, all of the wrapper storages.  I'm not sure how we do that.  Some I can think of:

- ZEO
- ZRS
- zc.zlibstorage
- zc.beforestorage

Those are ones I'm on the hook for.

I have a feeling that this is going to get bloody. :)

I'm not sure I want to make this required in ZODB 5.  There's just so much else in motion.

jmuchemb:
I already thought about zc.zlibstorage and it does not need any change: it forwards the returned value of `store` verbatim, and it does not hook `tpc_vote`/`tpc_finish`.
I didn't know zc.beforestorage but obviously, there's also nothing to do here.

> I have a feeling that this is going to get bloody. :)

Upgrading to the new API is simple and this has 0 impact on data. Even the change in [BTrees to forbit None keys](/zopefoundation/ZODB/commit/bb5aac21277f43333d6450064dc6670c8c280e40) was more problematic ([to use ERP5 with ZODB4](https://lab.nexedi.com/nexedi/erp5/merge_requests/112)).

We can write notes to describe how to upgrade. In fact, that would be an english translation of the [DemoStorage monkey-patch in testDemoStorage](https://github.com/zopefoundation/ZODB/blob/caea03caf518f243370ac5cc01febdba6a7d701d/src/ZODB/tests/testDemoStorage.py#L47).

> I'm not sure I want to make this required in ZODB 5.

That could affect the warning I wrote.

jimfulton:
I'm not opposed to making empty threats. ;)

Thanks for checking zlibstorage and beforestorage.

jmuchemb:
ok, so we can still decide in the PR I'll do for the ZODB 5.

(Should I wait for someone else to review ? I need this merged to continue, but that can wait tomorrow since it's already late for me.)

jimfulton:
This has been released as ZODB 4.4.0

jimfulton:
And merged to master

jimfulton:
A complication for ZEO, is that it can't support clients that use the old protocol unless the storages it wraps support the old protocol. What's worse, there's no good way for ZEO to know how a storage will behave until it tries to commit a transaction.  I think the interfaces need to be adjusted to allow this introspection. I'll make another PR.

PR #71, opened 2016-06-17
Move load to imvcc storage

jimfulton:
This relatively small cleanup PR moves load to IMVCCStorage from IStorage.

It also loses the version argument from load and from IStorageWrapper.invalidate.


jamadden:
LGTM. Travis is green.

PR #72, opened 2016-06-18
Modernize and cleanup locking

jimfulton:
- Move from try/finally to with

- Start phasing out the _lock_acquire/_lock_release shortcuts

- Replace simpler @locked decorators with with statements.

  I've come to prefer the with style tp the decorator style.
  I think it's clearer.

  (The decorators preceeded the Python with statement.)

  I left the decorators in cases where they were used with preconditions.

This is mostly pretty mechanical, although it got a little delicate in
places and ... tests

### DIFF
--- a/src/ZODB/ActivityMonitor.py
+++ b/src/ZODB/ActivityMonitor.py
@@ -41,18 +41,15 @@ class ActivityMonitor:
         self.trim(now)
 
     def trim(self, now):
-        self.trim_lock.acquire()
-
-        log = self.log
-        cutoff = now - self.history_length
-        n = 0
-        loglen = len(log)
-        while n < loglen and log[n][0] < cutoff:
-            n = n + 1
-        if n:
-            del log[:n]
-
-        self.trim_lock.release()
+        with self.trim_lock:
+            log = self.log
+            cutoff = now - self.history_length
+            n = 0
+            loglen = len(log)
+            while n < loglen and log[n][0] < cutoff:
+                n = n + 1
+            if n:
+                del log[:n]
 
     def setHistoryLength(self, history_length):
         self.history_length = history_length
--- a/src/ZODB/BaseStorage.py
+++ b/src/ZODB/BaseStorage.py
@@ -85,13 +85,13 @@ class BaseStorage(UndoLogCompatible):
 
         # Allocate locks:
         self._lock = utils.RLock()
-        self.__commit_lock = utils.Lock()
+        self._commit_lock = utils.Lock()
 
-        # Comment out the following 4 lines to debug locking:
+        # Needed by external storages that use this dumb api :(
         self._lock_acquire = self._lock.acquire
         self._lock_release = self._lock.release
-        self._commit_lock_acquire = self.__commit_lock.acquire
-        self._commit_lock_release = self.__commit_lock.release
+        self._commit_lock_acquire = self._commit_lock.acquire
+        self._commit_lock_release = self._commit_lock.release
 
         t = time.time()
         t = self._ts = TimeStamp(*(time.gmtime(t)[:5] + (t%60,)))
@@ -128,8 +128,8 @@ class BaseStorage(UndoLogCompatible):
     def new_oid(self):
         if self._is_read_only:
             raise POSException.ReadOnlyError()
-        self._lock_acquire()
-        try:
+
+        with self._lock:
             last = self._oid
             d = byte_ord(last[-1])
             if d < 255:  # fast path for the usual case
@@ -139,19 +139,14 @@ class BaseStorage(UndoLogCompatible):
                 last = _structpack(">Q", last_as_long + 1)
             self._oid = last
             return last
-        finally:
-            self._lock_release()
 
     # Update the maximum oid in use, under protection of a lock.  The
     # maximum-in-use attribute is changed only if possible_new_max_oid is
     # larger than its current value.
     def set_max_oid(self, possible_new_max_oid):
-        self._lock_acquire()
-        try:
+        with self._lock:
             if possible_new_max_oid > self._oid:
                 self._oid = possible_new_max_oid
-        finally:
-            self._lock_release()
 
     def registerDB(self, db):
         pass # we don't care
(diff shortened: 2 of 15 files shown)


src/ZODB/BaseStorage.py
```diff
@@ -85,13 +85,13 @@ def __init__(self, name, base=None):
 
         # Allocate locks:
         self._lock = utils.RLock()
-        self.__commit_lock = utils.Lock()
```
tseaver: Yay!

src/ZODB/ActivityMonitor.py
```diff
@@ -41,18 +41,15 @@ def closedConnection(self, conn):
...
-        self.trim_lock.release()
+        with self.trim_lock:
+            log = self.log
+            cutoff = now - self.history_length
+            n = 0
+            loglen = len(log)
+            while n < loglen and log[n][0] < cutoff:
+                n = n + 1
+            if n:
+                del log[:n]
```
tseaver: Fixes a deadlock if any of that code could ever raise.

src/ZODB/BaseStorage.py
```diff
@@ -245,8 +235,7 @@ def tpc_finish(self, transaction, f=None):
         # to send an invalidation message to all of the other
         # connections!
 
-        self._lock_acquire()
-        try:
+        with self._lock:
             if transaction is not self._transaction:
                 raise POSException.StorageTransactionError(
                     "tpc_finish called with wrong transaction")
```
tseaver: When / how does `self._commit_lock` get released in this case?

jimfulton: It gets released in the finally below.

(I may be misunderstanding your question.)

src/ZODB/FileStorage/fspack.py
```diff
@@ -445,11 +442,10 @@ def close_files_remove():
             # pack didn't free any data.  there's no point in continuing.
             close_files_remove()
             return None
-        self._commit_lock_acquire()
+        self._commit_lock.acquire()
```
tseaver: Since acquire/release all occur in the same method, wouldn't this be clearer using `with self._commit_lock:`?

src/ZODB/MappingStorage.py
```diff
@@ -269,9 +267,9 @@ def tpc_begin(self, transaction, tid=None):
         if transaction is self._transaction:
             raise ZODB.POSException.StorageTransactionError(
                 "Duplicate tpc_begin calls for same transaction")
-        self._lock_release()
+        self._lock.release()
         self._commit_lock.acquire()
-        self._lock_acquire()
+        self._lock.acquire()
```
tseaver: Maybe the `utils.RLock` needs to grow a second context manager helper, to handle a scoped release-then-acquire.

src/ZODB/FileStorage/fspack.py
jimfulton: <shrug>

This packing code is very old and really ought to be updated.  But this is out of scope of this PR.

src/ZODB/MappingStorage.py
jimfulton: ``utils.RLock`` is only used when debugging locks. The locks used here are regular locks normally.

This would be a little clearer if I didn't use the decorator and used 2 with's.  I decided to still with the decorators, as they're used pretty consistently in this class.

tseaver: What if we add a helper class, something like:

```python
import contextlib


@contextlib.contextmanager
def LockReleaser(real_lock):
    """Context manager:  release a lock temporarily.
    
    E.g.:

        with real_lock:
            do_some_work_with_lock()
            with LockReleaser(real_lock):
                do_work_without_lock()
            do_more_work_with_lock()
    """
    real_lock.release()
    try:
        yield None
    finally:
        real_lock.acquire()
```

jimfulton: I addressed this another way.

jimfulton:
@tseaver are you good with this now?

src/ZODB/BaseStorage.py
```diff
@@ -85,13 +85,13 @@ def __init__(self, name, base=None):
...
+        self._commit_lock = utils.Lock()
 
-        # Comment out the following 4 lines to debug locking:
+        # Needed by external storages that use this dumb api :(
         self._lock_acquire = self._lock.acquire
         self._lock_release = self._lock.release
-        self._commit_lock_acquire = self.__commit_lock.acquire
-        self._commit_lock_release = self.__commit_lock.release
+        self._commit_lock_acquire = self._commit_lock.acquire
+        self._commit_lock_release = self._commit_lock.release
```
tseaver: Are these micro-optimizations still worth preserving?

tseaver:
OK to merge even without addressing my release-then-acquire concerns (it still smells like a bug magnet, though).

src/ZODB/BaseStorage.py
jimfulton: They're part of the old inheritance API. :( The comment above explains this.

jimfulton:
I thought I *did* address your concerns. Did you see the change I made?  The locked regions for the ordinary lock should be very obvious now.

src/ZODB/BaseStorage.py
tseaver: Ugh, what a terrible pattern:  who could have done such a thing? (he asks innocently).

tseaver:
The one that I commented on last me  the `_commit_lock`, which is [acquired in one method](https://github.com/zopefoundation/ZODB/pull/72/files#diff-71117ab239f0de5734a0a100e675fe14R183), held across multiple method invocations, and released later.  It is also [released-then-acquired two layers down](https://github.com/zopefoundation/ZODB/pull/72/files#diff-808cb96d0cd7582ec5350a774ad64854R637) in subclass code.

jimfulton:
OK, wrt commit lock being held over multiple methods:

a) This is a key aspect of the two-phase commit protocol.  At least when we return from voting, we have to be assured that we can commit the modified objects, so we have to prevent modification while in the second phase.  Currently, we get the lock in tpc_begin.  This is a design flaw.  We don't really need to get the vote until tpc_vote, and of course, we should be able to get object-level locks, rather than a global lock.  This is related to https://github.com/zopefoundation/ZODB/pull/69.  So, eventually we won't get locks until tpc_vote, but even then, the lock will be acquired in tpc_vote and released in tpc_finish or tpc_abort.

b) This is out of scope of this change. :)  Although, I think using ``with``, especially when temporarily releasing the storage lock to acquire the commit lock, makes the logic a little clearer.

PR #76, opened 2016-06-30
Define a new interface for storages that support the new commit protocol.

jimfulton:
This allows interface incompatibilities:

- Future version of ZODB that requires the new protocol being
  configured with a storage that doesn't implement the new protocol

- ZEO has a client that uses the old protocol can't use a storage that
  implements the new protocol.

To be caught early.

This commit also clarifies that the serial returned by ``undo()`` is ignored.


jimfulton:
@jmuchemb and @jamadden please take a look at this

jamadden:
Perhaps the base methods in `IStorage`, like `tpc_vote`, should drop the new language about return types that was added in #69 and just mention `IMultiCommitStorage` instead? That might make it a bit less confusing. Since it looks like none of the built-in implementations of IStorage, like FileStorage, started doing the new thing yet, that should only be a docs change for now.

It also might make it less confusing if the docs for `IMultiCommitStorage` didn't *almost* duplicate the docs from `IStorage`. That makes it difficult to see what the differences are at a glance. Perhaps they could be something like:
```python
def tpc_vote(...<real params here>):
    """
     Accepts the same arguments and has the same error conditions
     as IStorage.tpc_vote; the only difference is the return value.

   <describe the return value>
   """
```

And then, just to be clear, RelStorage would `@implementer(IMVCCStorage,IMultiCommitStorage)` when it adopts this.

jimfulton:
The goal of https://github.com/zopefoundation/ZODB/pull/69 was to allow storages to implement the new interface, while supporting storages that implemented the old. Updating the built-in storages would have broken ZEO.

jimfulton:
In order for IMultiCommitStorage to extend IStorage, it should conform to IStorage, which is why the allowable return values in IStorage were expanded.  I don't *think* you're objecting to this.

I'll work on the wording a bit to try to reduce the duplication and make the distinction clearer.

jamadden:
> In order for IMultiCommitStorage to extend IStorage, it should conform to IStorage, which is why the allowable return values in IStorage were expanded. I don't think you're objecting to this.

No, not at all. Just the duplication and the fact that the IStorage docs become rather difficult to parse given all the conditionals. Since there's a new interface that defines the expected return types for the new protocol, if it were me, I would have the IStorage docs say something like:
```python
def tpc_vote(...):
   """
   ...
   Returns <whatever it returns>.

   If the return value is not as specified here, it is allowed if and only if the implementation 
   provides a subclass of IStorage that further specifies the return value.
   """
```

But that's me. Others may still prefer having the root interface explicitly list everything that all its subclasses support too.

src/ZODB/interfaces.py
```diff
@@ -797,104 +793,48 @@ def tpc_vote(transaction):
...
-
-        StorageTransactionError
-          is raised when transaction does not match the current
-          transaction.
-
-        StorageError or, more often, a subclass of it
-          is raised when an internal error occurs while the storage is
-          handling the store() call.
-
+        See IStorage.store. For object's implementing this interface,
```
jamadden: No apostrophe in "object's".

jamadden:
That reads a lot better to me, thank you. 👍

jmuchemb:
I'm not sure to understand the purpose of this PR. Incompatibility can be checked at store/tpc_vote: is this to fail earlier (like `DB.__init__`) when we'll drop support for the old API ? Who would upgrade ZODB without even testing a single write?

It may also better to finish what I've started and posted the zodb-dev ([New API to update _p_changed/_p_serial after a commit](http://thread.gmane.org/gmane.comp.web.zope.zodb/13267)), so that we get a clear view of the current status. The fact is there is pending work in master to update all storages to the new API.

jimfulton:
Yes, this is to fail sooner.  It would never occur to me to "test" an upgrade specifically by writing.

Even with this, clients may fail at connect time, which may not be an obvious error, but I don't see a way to do better.  A potential added benefit is that it could allow Connection to branch one earlier by configuring itself to use simpler logic when a storage provides ``IMultiCommitStorage``.

jimfulton:
ZODB has a very modular architecture. This is both a strength and a weakness.  It's great that parts can be swapped out and improved, but a major downside is that it can be difficult for new or casual users to know that the parts are assembled correctly.  Failing sooner and more obviously can help. Also having interfaces that describe behavior more explicitly will also help during the transition I think.

PR #77, opened 2016-07-01
Better support of the new API to notify of resolved conflicts (store/tpc_finish)

jmuchemb:


jmuchemb:
As mentioned on zodb-dev, I splitted what I did in 2 commits: the first one is this PR, to fix issues in the 4.x branch.
To better understand this PR, you may want to look at the branch handle-serial, which contains the next part for ZODB5: 1 merge commit to bring this PR + 4558d9398b69c54950c233dc981607939956e111 to switch all storages to the new API.

jimfulton:
Could you briefly describe the the issues this is addressing?

jmuchemb:
The only commit of this PR has a message which, I think, describes everything. Is there anything missing ?

Maybe I can add that it was decided that ZODB4 would support both API, so that storages can already move to the new one, and this PR fixes a few issues in the new API that weren't known before because NEO does not use blobs.

gotcha:
Copy Paste of the commit message:

## Better support of the new API to notify of resolved conflicts (store/tpc_finish)

When switching all storages to the new API in the master branch,
I found a few issues, mainly with blobs.

1. Created/modified blobs are invalidated during the first phase
   (in Connection._store_objects):

    obj._p_invalidate()

    And with the old API, _handle_serial then sets _p_serial
    Ghost objects are not supposed to have a _p_serial and with the new API,
    a few tests would fail because _p_serial are checked without activating the
    blob first.

2. Another consequence of _handle_serial not updating _p_changed/_p_serial
   immediately is that created objects are stored twice if __getstate__
   modifies itself. This case is tested in testConnection by
   doctest_lp485456_setattr_in_setstate_doesnt_cause_multiple_stores

    Hence the change in Connection._commit:
    - self._modified is already appended in Connection._store_objects
    - (obj._p_serial == z64) instead of (oid in self._creating)
      would not work for savepoints.

3. Setting _p_changed of a Blob with no uncommitted changes would cause an error
   (lp440234_Setting__p_changed_of_a_Blob_w_no_uncomitted_changes_is_noop)

    Fixed by the same change as in 2:
    - oid was appended twice to self._modified but reverted one
      (self._modified.pop() in _store_objects)
    - the test passed because _p_changed was reset early by _handle_serial

jimfulton:
OK, thanks for pointing out the commit message.  So you're arguing that the blob issue was with the tests. I accept that.

Something I regret about the original PR that is still a problem with this one is that changes aren't tested. (I wish I'd been a better reviewer)  I discovered this the hard way when I tried to update ZEO to support the new protocol.

See https://github.com/zopefoundation/ZODB/pull/80, which augments this PR with tests, additional test fixes, and undo fixes.

jmuchemb:
The commit I've pushed is similar to 6035691504667cf604dd70bfaa2af5d31a85bd8c from #79 
The difference is that I keep a change I did previously to get _p_serial after being sure the oid is not ghost.

PR #79, opened 2016-07-02
Various fixes for the new commit protocol (IMultiCommitStorage)

jimfulton:
When I started updating ZEO to provide IMultiCommitStoage, I ended up with many shared test failures, some clearly due to ZODB bugs.

I realized that ZODB hadn't actually been tested with IMultiCommitStorage storages, so I wrote a storage adapter and caused many file-storage tests to be run with the adapter.  This exposed a number of bugs, including non-test bugs, which are fixed by this PR.


jimfulton:
@jmuchemb This is relevant to the work you're doing.

Note that I needed these fixes to make progress on updating ZEO to provide IMultiCommitStorage, which is done, bit awaiting a ZODB release with these fixes to make a PR.

@jamadden likely want to see these too, assuming that you intend to update RelStorage soon.

jimfulton:
Gaaa, I wrote this using Python 2.7.  Fixing for Py3...

jmuchemb:
I disagree with changes that are related to blobs (in particular the change in tpc_finish). I agree there are issues to fix but not this way, and I did it already in #77 and the handle_serial branch.

In the handle_serial branch, I updated src/ZODB/tests/IExternalGC.test in [a stricter way](/zopefoundation/ZODB/commit/4558d9398b69c54950c233dc981607939956e111#diff-bde18894783113cdcda894da7542d31b)

For the undo part, there are indeed issues I haven't noticed. Which means there lack tests.

Can we postpone the adapter part, since it currently conflicts with the fact that I switch all storages to the new API ? (still in the handle_serial branch)

jmuchemb:
Concerning undo, what about ignoring the value returned by `tpc_vote` ?
I don't see why `undo()` would not return all oids to be invalidated.

jimfulton:
See my comments on the handle-serial branch and my fixes to it.

"For the undo part, there are indeed issues I haven't noticed. Which means there lack tests."

Right, your PRs never tested against a storage that actually implemented the new commit protocol.  Had you done that, then a number of existing tests would have failed.  See https://github.com/zopefoundation/ZODB/pull/80.

The adapter can go away later. For now, it enables running existing tests against storages that fully implement the new protocol.

WRT undo, yes undo *could* return the oids, however, ZEO has implemented undo asynchronously for some time and it's impractical to change this now.  There are existing ZODB undo tests that specifically allow for oids to be returned from undo via vote. We could decide to change this later, but it's not clear that we should.

jimfulton:
I should note that I'm willing to defer to https://github.com/zopefoundation/ZODB/pull/77 and redo this based on that PR, assuming it incorporates https://github.com/zopefoundation/ZODB/pull/80.

I don't want to wait for storages to be updated to implement the new protocol, unless that happens in the next day or 2 (or perhaps 3), as this is blocking ZEO work.

PR #80, opened 2016-07-04
Test that IMultiCommitStorages are supported

jimfulton:
This makes sure that there is support for the new protocol by testing against a storage that supports the new protocol.

Note that ZEO has implemented undo asynchronously and ZODB has supported undone OIDs to be returned via ``tpc_vote`` for for some time. It could be argued whether this is a good thing or a bad thing. (My opinion is that it's simpler to return all affected oids in the vote result, but I don't feel strongly about it.)  It is, however, impractical to change this now.

### DIFF
--- a/src/ZODB/DB.py
+++ b/src/ZODB/DB.py
@@ -1012,8 +1012,8 @@ class TransactionalUndo(object):
     def tpc_vote(self, transaction):
         result = self._storage.tpc_vote(transaction)
         if result:
-            if isinstance(result[0], bytes):
-                self._oids.update(set(result))
+            if isinstance(next(iter(result)), bytes):
+                self._oids.update(result)
             else:
                 for oid, _ in result:
                     self._oids.add(oid)
--- a/src/ZODB/DB.py
+++ b/src/ZODB/DB.py
@@ -1010,8 +1010,13 @@ class TransactionalUndo(object):
                 self._oids.update(result[1])
 
     def tpc_vote(self, transaction):
-        for oid, _ in  self._storage.tpc_vote(transaction) or ():
-            self._oids.add(oid)
+        result = self._storage.tpc_vote(transaction)
+        if result:
+            if isinstance(result[0], bytes):
+                self._oids.update(set(result))
+            else:
+                for oid, _ in result:
+                    self._oids.add(oid)
 
     def tpc_finish(self, transaction):
         self._storage.tpc_finish(
--- a/src/ZODB/multicommitadapter.py
+++ b/src/ZODB/multicommitadapter.py
@@ -11,7 +11,7 @@ class MultiCommitAdapter:
         self._storage = storage
         ifaces = zope.interface.providedBy(storage)
         zope.interface.alsoProvides(self, ifaces)
-        self._resolved = set()
+        self._resolved = set() # {OID}, here to make linters happy
 
     def __getattr__(self, name):
         v = getattr(self._storage, name)
@@ -39,7 +39,7 @@ class MultiCommitAdapter:
     def undo(self, transaction_id, transaction):
         r = self._storage.undo(transaction_id, transaction)
         if r:
-            self._resolved.update(set(r[1]))
+            self._resolved.update(r[1])
 
     def tpc_vote(self, *args):
         s = self._storage.tpc_vote(*args)
@@ -47,7 +47,7 @@ class MultiCommitAdapter:
             if serial == ResolvedSerial:
                 self._resolved.add(oid)
 
-        return list(self._resolved)
+        return self._resolved
 
     def tpc_finish(self, transaction, f=lambda tid: None):
 
--- a/src/ZODB/multicommitadapter.py
+++ b/src/ZODB/multicommitadapter.py
@@ -1,0 +1,65 @@
+"""Adapt non-IMultiCommitStorage storages to IMultiCommitStorage
+"""
+
+import zope.interface
+
+from .ConflictResolution import ResolvedSerial
+
+class MultiCommitAdapter:
+
+    def __init__(self, storage):
+        self._storage = storage
+        ifaces = zope.interface.providedBy(storage)
+        zope.interface.alsoProvides(self, ifaces)
+        self._resolved = set()
+
+    def __getattr__(self, name):
+        v = getattr(self._storage, name)
+        self.__dict__[name] = v
+        return v
+
+    def tpc_begin(self, *args):
+        self._storage.tpc_begin(*args)
+        self._resolved = set()
+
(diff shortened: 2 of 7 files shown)


jmuchemb:
About undo, I have mixed feelings. On one side, I don't like to support several API (oids returned either from undo or from tpc_vote) without benefit, and I would prefer ZEO to comply with the current interface. On the other site, it's only a matter of 2-3 lines, that are not performance critical.

The new commit protocol is already tested in testDemoStorage. But that does not test blobs.

So globally, I'm ok with this PR. Now commenting the diff...

src/ZODB/DB.py
```diff
@@ -1010,8 +1010,13 @@ def commit(self, transaction):
                 self._oids.update(result[1])
 
     def tpc_vote(self, transaction):
-        for oid, _ in  self._storage.tpc_vote(transaction) or ():
-            self._oids.add(oid)
+        result = self._storage.tpc_vote(transaction)
+        if result:
+            if isinstance(result[0], bytes):
+                self._oids.update(set(result))
```
jmuchemb: no need to cast as `set`

src/ZODB/multicommitadapter.py
```diff
@@ -0,0 +1,65 @@
...
+
+from .ConflictResolution import ResolvedSerial
+
+class MultiCommitAdapter:
+
+    def __init__(self, storage):
+        self._storage = storage
+        ifaces = zope.interface.providedBy(storage)
+        zope.interface.alsoProvides(self, ifaces)
+        self._resolved = set()
```
jmuchemb: Isn't the line in tpc_begin enough ?

src/ZODB/multicommitadapter.py
```diff
@@ -0,0 +1,65 @@
...
+                s = ((oid, s), )
+
+            for oid, serial in s:
+                if s == ResolvedSerial:
+                    self._resolved.add(oid)
+
+    def undo(self, transaction_id, transaction):
+        r = self._storage.undo(transaction_id, transaction)
+        if r:
+            self._resolved.update(set(r[1]))
```
jmuchemb: No need to cast as set

src/ZODB/multicommitadapter.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def tpc_vote(self, *args):
+        s = self._storage.tpc_vote(*args)
+        for (oid, serial) in (s or ()):
+            if serial == ResolvedSerial:
+                self._resolved.add(oid)
+
+        return list(self._resolved)
```
jmuchemb: no need to cast as `list`

src/ZODB/tests/blob_transaction.txt
```diff
@@ -402,14 +398,7 @@ Now, the serial for the existing blob should be the same:
...
     True
 
-And we shouldn't be able to read the data that we saved:
-
-    >>> blob_storage.loadBlob(blob._p_oid, s1)
-    Traceback (most recent call last):
-    ...
-    POSKeyError: 'No blob file at <BLOB STORAGE PATH>'
-
-Of course the old data should be unaffected:
```
jmuchemb: I could keep this in the handle_serial branch: https://github.com/zopefoundation/ZODB/commit/4558d9398b69c54950c233dc981607939956e111#diff-f3c7ea79af896e7fe5d1dfea68a004fcR403

src/ZODB/multicommitadapter.py
jimfulton: It is, but some reviewers' linters complain if variables aren't initialized, and I prefered here to initialize with a valid value.

jmuchemb: Then, a comment on such useless lines would be good (btw, what about `None` instead of `set()`). It's horrible to ignore the purpose of a line when refactoring code.

src/ZODB/tests/blob_transaction.txt
jimfulton: Because you didn't test it.  storeBlob doesn't return s1, not does tpc_vote, because there wasn't a resolved conflict.

jmuchemb: It's true I hacked a little to get the tid, but I thought the intention was to also check that everything was reverted with tpc_abort.

jmuchemb: There is also
```
 >>> blob_storage.loadBlob(blob._p_oid, s2)
```
below.

src/ZODB/multicommitadapter.py
jimfulton: It needed to be a list because TransactionalUndo and test callers looked at the first element to decide if it's a collection of oids or a collection of (oid, serial) pairs.  I'm not a fan of these sorts of conversions, or of assuming types when I don't need to, so I'm changing the callers to use ``next(iter(result))``.

jmuchemb: Oops, I missed that. OK.

Note:
```
$ python -m timeit -s 'x = set("bar")' 'next(iter(x))'
10000000 loops, best of 3: 0.169 usec per loop
$ python -m timeit -s 'x = list("bar")' 'x[0]'
10000000 loops, best of 3: 0.0282 usec per loop
```
So for me, forcing storages to return a list could be fine if there's no benefit for storages to use something else than a list.

Hopefully, this test is temporary.

src/ZODB/tests/blob_transaction.txt
jimfulton: Oh gawd, the machinery that loads blobs doesn't mind being called with None for the tid, so the last lines I removed above are meaningless. If you want to add a tid hack, be my guest, but I don't think it's necessary. I think the check:

     >>> blob_storage.load(blob._p_oid, '') == (olddata, oldserial)
     True

is sufficient.

jmuchemb: ok, let's delete

src/ZODB/multicommitadapter.py
jimfulton: Too late. :)

Anyway, when we get rid of the test, which *is* temporary, then we'll be glad we didn't have to create a list.

src/ZODB/DB.py
jimfulton: I didn't think so, but I failed with this the other day. (The think I know now what my faulty flail was.) Fixed, and also below.

jimfulton:
OK, I think I've made the needed changes. Waiting for travis...

jimfulton:
and travis is green

jmuchemb:
I merged. In any case, discussion can continue on #77

PR #82, opened 2016-07-04
Switch all storages to the new commit protocol

jmuchemb:
This does not clean up tests. In particular, the test using MultiCommitAdapter becomes useless (and IMO, MultiCommitAdapter too). Several changes from 74fba08ad4dc7311ca74c5c9da51c12e0e9dca86 could be also simplified.

### DIFF
--- a/src/ZODB/DemoStorage.py
+++ b/src/ZODB/DemoStorage.py
@@ -73,6 +73,7 @@ class DemoStorage(ConflictResolvingStorage):
 
         self._issued_oids = set()
         self._stored_oids = set()
+        self._resolved = []
 
         self._commit_lock = ZODB.utils.Lock()
         self._transaction = None
@@ -116,7 +117,7 @@ class DemoStorage(ConflictResolvingStorage):
         for meth in (
             '_lock',
             'getSize', 'isReadOnly',
-            'sortKey', 'tpc_transaction', 'tpc_vote',
+            'sortKey', 'tpc_transaction',
             ):
             setattr(self, meth, getattr(changes, meth))
 
@@ -308,7 +309,8 @@ class DemoStorage(ConflictResolvingStorage):
 
         if old != serial:
             rdata = self.tryToResolveConflict(oid, old, serial, data)
-            self.changes.store(oid, old, rdata, '', transaction, True)
+            self.changes.store(oid, old, rdata, '', transaction)
+            self._resolved.append(oid)
         else:
             self.changes.store(oid, serial, data, '', transaction)
 
@@ -364,6 +366,13 @@ class DemoStorage(ConflictResolvingStorage):
             self.changes.tpc_begin(transaction, *a, **k)
             self._transaction = transaction
             self._stored_oids = set()
+            del self._resolved[:]
+
+    def tpc_vote(self, *a, **k):
+        if self.changes.tpc_vote(*a, **k):
+            raise ZODB.POSException.StorageTransactionError(
+                "Unexpected resolved conflicts")
+        return self._resolved
 
     def tpc_finish(self, transaction, func = lambda tid: None):
         with self._lock:
--- a/src/ZODB/DemoStorage.py
+++ b/src/ZODB/DemoStorage.py
@@ -32,7 +32,7 @@ import ZODB.POSException
 import ZODB.utils
 import zope.interface
 
-from .ConflictResolution import ConflictResolvingStorage, ResolvedSerial
+from .ConflictResolution import ConflictResolvingStorage
 from .utils import load_current, maxtid
 
 @zope.interface.implementer(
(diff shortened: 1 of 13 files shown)


src/ZODB/DemoStorage.py
```diff
@@ -309,7 +309,7 @@ def store(self, oid, serial, data, version, transaction):
         if old != serial:
             rdata = self.tryToResolveConflict(oid, old, serial, data)
             self.changes.store(oid, old, rdata, '', transaction)
-            return ResolvedSerial
+            return True
```
jimfulton: Why is this store here? (Or the one below?)

Return values from store are no longer used.

jimfulton: Disregard the comment above.  But we need to correct ``store``'s contract in IMultiCommitStorage.

jmuchemb:
I refreshed the PR:
- 1 merge commit to bring changes from the 4.x branch: there were conflicts in the changelog
- the original commit of this PR that is rebased

This time, I've removed the tests using MultiCommitAdapter: I had test failures I didn't want to solve because these tests become redundant.

src/ZODB/tests/blob_transaction.txt
```diff
@@ -381,16 +381,17 @@ stored are discarded.
...
     ...     _ = file.write(b'This data should go away too')
-    >>> s2 = blob_storage.storeBlob(new_oid, '\0'*8, olddata, 'blobfile2',
+    >>> blob_storage.storeBlob(new_oid, '\0'*8, olddata, 'blobfile2',
     ...                             '', t)
-
-    >>> serials = blob_storage.tpc_vote(t)
-
+    >>> bool(blob_storage.tpc_vote(t))
+    False
+    >>> oldserial < blob_storage._tid
```
jimfulton: You're using underware!

This fails for ZEO.

What is this trying to test?

src/ZODB/FileStorage/FileStorage.py
```diff
@@ -536,9 +535,7 @@ def store(self, oid, oldserial, data, version, transaction):
                     "The storage quota has been exceeded.")
 
             if old and oldserial != committed_tid:
-                return ResolvedSerial
-            else:
-                return self._tid
+                self._resolved.append(oid)
```
jimfulton: Move this line after the the call to ``self.tryToResolveConflict`` and get rid of this ``if``.

src/ZODB/DemoStorage.py
```diff
@@ -308,10 +308,9 @@ def store(self, oid, serial, data, version, transaction):
 
         if old != serial:
             rdata = self.tryToResolveConflict(oid, old, serial, data)
-            self.changes.store(oid, old, rdata, '', transaction)
-            return ResolvedSerial
-
-        return self.changes.store(oid, serial, data, '', transaction)
+            self.changes.store(oid, old, rdata, '', transaction, True)
```
jimfulton: ``self.changes`` could be anything. You can't assume it takes the resolved argument.

Why not let ``DemoStorage`` have its own ``_resolved``?

src/ZODB/MappingStorage.py
```diff
@@ -232,7 +233,7 @@ def sortKey(self):
 
     # ZODB.interfaces.IStorage
     @ZODB.utils.locked(opened)
-    def store(self, oid, serial, data, version, transaction):
+    def store(self, oid, serial, data, version, transaction, resolved=False):
```
jimfulton: Please no the ``resolved`` argument. :)  It couples ``MappingStorage`` and ``DemoStorage`` too tightly.

Also, one of my goals for ``MappingStorage`` was that it would provide an illustration of how the storage API might be implemented.

``DemoStorage`` does conflict resolution.  ``MappingStorage`` doesn't and should have ``_resolved``.

jmuchemb: > `MappingStorage` doesn't and should have `_resolved`.

Do you mean it shouldn't ?

I'll move all `_resolved` in  `MappingStorage` to `DemoStorage`.

src/ZODB/tests/blob_transaction.txt
jmuchemb: Indeed, I can't do that.
The test was originally checking things with the tid being committed (which was known by the returned value of `store`). I tried to keep that.
Since I wrote this change, we already decided to delete some lines after this part.

I'm going to delete.

src/ZODB/FileStorage/FileStorage.py
jmuchemb: :)

src/ZODB/MappingStorage.py
jimfulton: Yes, sorry, it shouldn't.

Thanks.

jimfulton:
Released as 5.0.0a5

PR #87, opened 2016-07-09
Per @jimfulton, handle_all_serials shouldn't be sniffing in ZODB5

jamadden:
So try that. This slightly simplifies _dostore.

There are lots of other callers still so I didn't remove the functions entirely (but I could try if desired).

### DIFF
--- a/src/ZODB/tests/StorageTestBase.py
+++ b/src/ZODB/tests/StorageTestBase.py
@@ -111,11 +111,8 @@ def handle_all_serials(oid, *args):
     This is pointless with IMultiCommitStorage.
     """
     d = {}
-    for arg in args:
-        if arg: # store() will have passed us None
-            for t in arg:
-                assert isinstance(t, bytes)
-                # This will be the tid returned by tpc_finish.
+    # On empty dict, _dostore will choose the serial returned from
+    # tpc_finish.
     return d
 
 def handle_serials(oid, *args):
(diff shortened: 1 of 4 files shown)


jamadden:
Turns out that at least the tests pass sequences of non-bytes to this method on Python 3 (probably str?), so I had to remove that assert.

src/ZODB/tests/StorageTestBase.py
```diff
@@ -194,10 +178,7 @@ def _dostore(self, oid=None, revid=None, data=None,
             r1 = self._storage.store(oid, revid, data, '', t)
```
jimfulton: We don't use r1 and r2, so why assign them?

jimfulton:
Gutting the methods is confusing.  Let's not leave this for a future code reader to stumble on. (It will probably be me and I just hurt myself for the last few minutes on this :) ).

We should remove these functions and their call points, or just remove the calls in dostore, if that's what was hurting RelStorage.

src/ZODB/tests/StorageTestBase.py
jamadden: I was trying to make the diff as minimal as possible. I can clean that up.

jamadden:
> On Jul 12, 2016, at 11:00, Jim Fulton <notifications@github.com> wrote:
> 
> Gutting the methods is confusing. Let's not leave this for a future code reader to stumble on. (It will probably be me and I just hurt myself for the last few minutes on this :) ).
> 
> We should remove these functions and their call points, or just remove the calls in dostore, if that's what was hurting RelStorage.

OK, I can remove the functions and their callers. That will involve touching three more files, but I think it's mostly copy-pasted code.

jamadden:
Travis appears to be green with the removal. I say appears because github has been lagging a lot lately.

PR #88, opened 2016-07-10
fstail: print the txn offset and header size, instead of only the data offset

jmuchemb:
Before:

    2016-07-01 09:41:50.416574: hash=d7101c5ee7b8e412d7b6d54873204421e09b7f34
    user='' description='' length=1629 offset=58990284

After:

    2016-07-01 09:41:50.416574: hash=d7101c5ee7b8e412d7b6d54873204421e09b7f34
    user='' description='' length=1629 offset=58990261 (+23)

The structure of a FileStorage DB is such that it's easy to revert the last
transactions, by truncating the file at the right offset. With the above
change, `fstail` can now be used to get this offset.

In the above example:

    truncate -s 58990261 Data.fs

would delete the transaction and all those after.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,9 @@
 - Internal FileStorage-undo fixes that should allow undo in some cases
   where it didn't work before.
 
+- fstail: print the txn offset and header size, instead of only the data offset.
+  fstail can now be used to truncate a DB at the right offset.
+
 4.4.2 (2016-07-08)
 ==================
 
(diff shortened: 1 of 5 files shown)


jamadden:
I like the idea of more attention and detail being paid to ZODB tools. I'm not an fs expert, but this looks basically reasonable to me.

jimfulton:
Could you add an entry to CHANGES.txt please?

PR #89, opened 2016-07-12
Refactored FileStorage transactional undo

jimfulton:
As part of a project to provide object-level commit locks for ZEO, I'm
refactiring FileStorage to maintain transaction-specific data in
Tranaction.data.  This involved undo.  In trying to figure this out, I
found:

- A bug in _undoDataInfo, which I verified with some tests and

- _transactionalUndoRecord was maddeningly difficult to reason about
  (and thus change).

I was concerned less by the bug than my inability to know whether a
change to the code would be correct.

So I refactored the code, mainly transactionalUndoRecord, to make the
code easier to understand, fixing some logic errors (I'm pretty sure)
along the way.  This included lots of comments. (Comments are much
easier to compose when you're working out logic you didn't
understand.)

In addition to makeing the code cleaner, it allows undo to be handled
in cases that weren't handled before.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,12 @@
  Change History
 ================
 
+4.4.3 (unreleased)
+==================
+
+- Internal FileStorage-undo fixes that should allow undo in some cases
+  where it didn't work before.
+
 4.4.2 (2016-07-08)
 ==================
 
--- a/src/ZODB/ConflictResolution.py
+++ b/src/ZODB/ConflictResolution.py
@@ -18,7 +18,8 @@ import six
 import zope.interface
 from ZODB.POSException import ConflictError
 from ZODB.loglevels import BLATHER
-from ZODB._compat import BytesIO, PersistentUnpickler, PersistentPickler, _protocol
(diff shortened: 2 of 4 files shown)


jimfulton:
Man, I reviewed 3 PRs today. Sure would be nice if this got reviewed. :)

jamadden:
That certainly seems a lot more clear to me; the comments are very helpful.

There are a lot of branches, I don't suppose you happened to compare coverage before/after? I assume the new test is coving more scenarios and hence more branches. (Might be nice if ZODB coverage was automatically reported to coveralls.io or some other service.)

The changes in src/ZODB/ConflictResolution.py seem unrelated unless I missed something.

jimfulton:
Yes, the changes in ConflictResolution.py are unrelated. I had to read that (and reread it) when working on this, and the long lines make my eyes bleed. That change was in the long-line conflict.

jimfulton:
The coverage on this branch is unchanged from the 4 branch at 79%.

jamadden:
Tests are green, code is easier to read and follow, so LGTM.

PR #90, opened 2016-07-12
Fix testMinimizeTerminates

jimfulton:
This test was spewing exceptions without failing for ZODB5.

The spew was an indication that something was wrong in this case. (The
test was broken.)

There had been some bug in the object cache that caused an infinite
loop when objects woke themselves up while being reaped.  This test is
a regression test for that bug.

The test was sloppy about transaction managers.  The test class
creates a bunch of connections using the threaded transaction manager,
but this test used a thread to execute the critical code.  As a
result, the synchronization methods weren't called.  This didn't
matter much with ZODB4.  The objects were still able to load their
state.

With ZODB5, the synchronization methods are important for proper MVCC
and because they weren't called, the objects' state couldn't be
reloaded, with 2 effects:

- Spew

- The test wasn't really testing that object's that reloaded
  themselves didn't create an infinite loop.

Changed the test to not use a threaded transaction manager.

### DIFF
--- a/src/ZODB/tests/testCache.py
+++ b/src/ZODB/tests/testCache.py
@@ -142,6 +142,13 @@ class DBMethods(CacheTestBase):
         # isn't looking out for this, it can get into an infinite loop
         # then, endlessly trying to ghostify an object that in turn keeps
         # unghostifying itself again.
+
+        # This test uses threads, so we can't use the default
+        # transaction manager.
+        for conn in self.conns:
+            conn.close()
+        self.conns[0] = self.db.open(transaction.TransactionManager())
+
         class Worker(threading.Thread):
 
             def __init__(self, testcase):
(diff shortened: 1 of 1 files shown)


jamadden:
The fix seems good to me.

But where can the exception spew be observed? I don't find it in a previous Travis build on master, say [build 381](https://travis-ci.org/zopefoundation/ZODB/builds/144225594). Is zope-testrunner hiding stderr output (and can that be changed if so)?

jimfulton:
I assumed it started when I started using loadBefore.  That's when not calling the synchronizers started to matter.

I certainly see it in build 381.

jamadden:
D'oh. Naturally the two environments I spot-checked, 2.7 and 3.5, don't demonstrate the problem. This is consistent across ad least two builds. How interesting.

PR #93, opened 2016-07-18
Drop old commit protocol

jmuchemb:
It was not tested anymore (since #82), and some clean-up was already done in #87.

Regarding my question on 870f2d58b795c5a29d67612037c60f40fb786cea, I don't change anything in this commit. The code still processes the value returned by tpc_vote when undoing transactions, even if for methods like `StorageTestBase._undo`, this is currently dead code.

Still about undo, there's an additional small change in `checkUndoMultipleConflictResolution` to check the undone value.

`Connection` simplifies a lot. No more `_handle_serial` method.

About interfaces, I suggest a few changes to the docstrings, but the result still seems weird to me. I understand that all storages should declare that they implement `IMultiCommitStorage` (not done), and that it should be checked (where?).


jmuchemb:
ok to merge ?

PR #95, opened 2016-07-23
Changed DB root-object-initialization to use an open connection and more

### LINKED ISSUE #84: DB root-object creation should use a connection
When a DB is created, it tried to make sure that the root object exists.  It doesn't use a connection to do this, but implements the transaction logic itself.  It hasn't kept up, however with recent changes in the MVCC implementation, as I discovered in some failing ZEO tests.  

DB logic should be updated to use a connection to do the transaction processing, even if changes are needed to Connection to deal with object 0 not being present.

jimfulton:
This fixes #84.

Also:

- Added missing transaction ``begin`` call in the DB ``transaction()``
  context manager.

- Added the ability to add a transaction note whan calling
  ``transaction()``.  This is useful (as would be the ability to pass
  in other transaction meta data, but this was needed by (and this
  tested) to make an existing test pass.

### DIFF
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -190,7 +190,6 @@ class Connection(ExportImport, object):
 
         self._reader = ObjectReader(self, self._cache, self._db.classFactory)
 
-
     def add(self, obj):
         """Add a new object 'obj' to the database and assign it an oid."""
         if self.opened is None:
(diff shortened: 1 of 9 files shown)


jamadden:
Setting aside the test failures, I really like the concept. It reads much cleaner to me. My main minor quibble is about the use of Connection underware (again, love that term!) by DB. But I guess they already had somewhat of an incestuous relationship, so not much change...

jimfulton:
Gaaaa PyPy test failures....... This is a *test* bug, which I'll chase down.

WRT underware, it's totally appropriate within an implementation and DB and Connection are in the same implementation.  The new private function is needed by DB to add the root object.  This shouldn't be a public API as other clients shouldn't be specify an OID.

The main goal of this isn't so much to make the code cleaner, although that is a benefit, but to avoid buggy duplication of code. The old implementation has a bug that I had to work around in a ZEO test.

PR #101, opened 2016-09-07
Reference docs

jimfulton:
Lot's of doc string changes and ZConfig updates to support generating reference documentation for zodb.org.

I've made a snapshot here: to facilitate review.

It's tempting to move the zodb docs into this repo to make reviewing changes easier.

### DIFF
--- a/src/ZODB/DB.py
+++ b/src/ZODB/DB.py
@@ -379,16 +379,16 @@ class DB(object):
              When a connection has remained unused in a connection
              pool for more than pool_timeout seconds, it will be
              discarded and it's resources released.
-        :param objects cache_size`: target maximum number of non-ghost
+        :param objects cache_size: target maximum number of non-ghost
              objects in each connection object cache.
-        :param int cache_size_bytes`: target total memory usage of non-ghost
+        :param int cache_size_bytes: target total memory usage of non-ghost
              objects in each connection object cache.
-        :param int historical_pool_size`: expected maximum number of total
+        :param int historical_pool_size: expected maximum number of total
             historical connections
-        :param objects historical_cache_size`: target maximum number
+        :param objects historical_cache_size: target maximum number
              of non-ghost objects in each historical connection object
              cache.
-        :param int historical_cache_size_bytes`: target total memory
+        :param int historical_cache_size_bytes: target total memory
              usage of non-ghost objects in each historical connection
              object cache.
         :param seconds historical_timeout: Maximum age of inactive
--- a/src/ZODB/DB.py
+++ b/src/ZODB/DB.py
@@ -39,7 +39,7 @@ import transaction
 from persistent.TimeStamp import TimeStamp
 import six
 
-from . import POSException
+from . import POSException, valuedoc
 
 logger = logging.getLogger('ZODB.DB')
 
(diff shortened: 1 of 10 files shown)


jimfulton:
See: https://github.com/zopefoundation/zodbdocs/pull/24

src/ZODB/interfaces.py
```diff
@@ -472,6 +471,9 @@ def close():
 
         Finalize the storage, releasing any external resources.  The
         storage should not be used after this method is called.
+
+        Note that daabses close theor storages when they're closed, so
```
icemac: Maybe `databases`?

src/ZODB/MappingStorage.py
```diff
@@ -32,8 +32,22 @@
         ZODB.interfaces.IStorageIteration,
         )
 class MappingStorage(object):
+    """In-memory storage implementation
+
+    Note that this implementation is somewhat naive and inefficient
+    with regard to locking.  It's implementation is primarily meant to
```
jamadden: *Its*

src/ZODB/interfaces.py
jamadden: Also *their*

src/ZODB/DB.py
```diff
@@ -385,23 +366,53 @@ def __init__(self, storage,
...
-          - `historical_cache_size_bytes` -- similar to `cache_size_bytes` for
-            the historical connection.
-          - `historical_timeout`: minimum number of seconds that
-            an unused historical connection will be kept, or None.
-          - `xrefs` - Boolian flag indicating whether implicit cross-database
-            references are allowed
+        :param objects historical_cache_size`: target maximum number
+             of non-ghost objects in each historical connection object
+             cache.
+        :param int historical_cache_size_bytes`: target total memory
```
jamadden: There's a stray backtick (`) at the end of several param names, leading to odd rendering.

src/ZODB/DB.py
```diff
@@ -385,23 +366,53 @@ def __init__(self, storage,
...
+             constructing cross-database references ans when accessing
+             database connections fron other databases.
+        :param dict databases: dictionary of database name to
+             databases in a multi-database configuration. The new
+             database will add itself to this dictionary. The
+             dictionary is used when getting connections in other databases.
+        :param boolean xrefs: Flag indicating whether cross-database
+            references are allowed from this database to other
+            databases in a multi-database configuration.
+        :param int large_record_size: When object records are saved
```
jamadden: Depending on your Sphinx theme, it *might* be useful to use `:keyword:` or equivalently `:key:` for the parameters that are usually passed, or are expected to be passed, as keywords (instead of `:param:`). Some themes render those slightly differently in a way that helps make that apparent.

src/ZODB/interfaces.py
jimfulton: Wow, that was a wreck. :)

src/ZODB/DB.py
jimfulton: ``key`` gave me a wildly different display that looked awful next to the ``param``. ``keyword`` didn't seem to have any effect, not even setting a class.

The documentation, http://www.sphinx-doc.org/en/stable/domains.html#python-signatures doesn't really mention differences between ``param`` alternatives.

I'm inclined to stick with ``param`` for now.

jamadden: It's *very* theme dependent. Some of them look better than others. Most of them don't seem to do anything. If it looks bad in your theme then so be it. Thanks for trying.

jimfulton:
OK, I've made the changes and updated the snapshot. (You may have to force refresh because caching.)

jimfulton:
Thanks for looking at this!!!

jamadden:
Looks pretty good to me!

PR #102, opened 2016-09-08
Merge zodbdocs

jimfulton:
This is an almost (there were some changes to README.rst) entirely mechanical update that merges the zodbdocs repository into this one.


jimfulton:
I've updated the travis config to make sure we can build html (not on pypy3) and also refactored to use buildout (toowtdi).

PR #104, opened 2016-09-08
Don't raise an AttributeError when DemoStorage successfully stores a blob first.

### LINKED ISSUE #103: 5.0.0. DemoStorage Exception
Hi,

In line 330 of DemoStorage.py (see below) we are not returning after we calling self.changes.storeBlob so the previous attribute error is always raised.

Thanks a lot,

Carlos

           if self._blobify():
                self.changes.storeBlob(
                    oid, oldserial, data, blobfilename, '', transaction)
            raise

jamadden:
Includes a (previously) failing test case.

Fixes #103.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,13 @@
  Change History
 ================
 
+5.0.1 (unreleased)
+==================
+
+- Fix an AttributeError that DemoStorage could raise if it was asked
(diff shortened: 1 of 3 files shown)


jamadden:
Build is green.

tseaver:
LGTM (even with grrrrr doctests).

PR #105, opened 2016-09-09
Documentation on transactions and threading.

jimfulton:
This adds documentatuon on transactions and threading concerns.

There are some topics, like application design and conflict resolution
that might want a deeper treatment, but would probably be better
handled through articles or dedicated topics. (I want to avoid
individual topics being too long or weedy to read, where practical.)

Writing this, I stumbled a bit over thread-local transaction managers.
For most applications, they don't add anything over accessing
transaction managers on connections and actually provide an
opportunity to fail. I'm convinced that it should be possible to do
most transaction management through connections and that the API
provided by transaction managers and the transaction package should be
reserved for distributed transactions.

I didn't mention gevent. I thik there should be a section on gevent,
but I think it should be written by someone who's used gevent with
ZODB. :)

Maybe there should also be a section or mention of using asyncio with
ZODB, pr maybe later.

Closes zopefoundation/zodbdocs#13
Closes zopefoundation/zodbdocs#16

(also a few minor ReST fixes)

### DIFF
--- a/doc/guide/transactions-and-threading.rst
+++ b/doc/guide/transactions-and-threading.rst
@@ -69,7 +69,7 @@ objects involved:
 
 Transaction
    Transactions represent units of work.  Each transaction has a beginning and
-   an end. Transaction provide the
+   an end. Transactions provide the
    :interface:`~transaction.interfaces.ITransaction` interface.
 
 Transaction manager
@@ -156,8 +156,8 @@ error.
 
 We used ``as trans`` above to get the transaction.
 
-Databases provide the :meth:`~ZODB.DB.transaction` to execute a code
-block in a transaction::
+Databases provide the :meth:`~ZODB.DB.transaction` method to execute a code
+block as a transaction::
 
   with db.transaction() as conn2:
      conn2.root.x += 1
@@ -166,10 +166,10 @@ block in a transaction::
 
    >>> exec(src)
 
-Here, when we used ``as``, we got a connection, not a transaction.
-This is because a new connection is opened by the
-:meth:`~ZODB.DB.transaction`` method.  A new transaction manager was
-used as well.
+This opens a connection, assignes it it's own context manager, and
+executes the nested code in a transaction.  We used ``as conn2`` to
+get the connection.  The transaction boundaries are defined ``with``
+statement.
 
 Getting a connection's transaction manager
 ------------------------------------------
@@ -206,9 +206,9 @@ same value for ``x`` that we set earlier:
   >>> conn.root.x
   3
 
-This is because it's still in the same transaction that was implicitly
-begun when a change was last committed against it.  If we want to see
-changes, we have to begin a new transaction:
+This is because it's still in the same transaction that was begun when
+a change was last committed against it.  If we want to see changes, we
+have to begin a new transaction:
 
   >>> trans = my_transaction_manager.begin()
   >>> conn.root.x
@@ -261,19 +261,73 @@ This isn't always easy.
 
 Sometimes you may need to queue some operations that update shared
 data structures, like indexes, so the updates can be made by a
-dedicated thread or process, without simultaneous updates.
+dedicated thread or process, without making simultaneous updates.
+
+Retrying transactions
+~~~~~~~~~~~~~~~~~~~~~
+
+The most common way to deal with conflict errors is to catch them and
+retry transactions.  To do this manually, involves code that looks
+something like this::
+
+  max_attempts = 3
+  attempts = 0
+  while 1:
+      try:
+          with transaction.manager:
+              ... code that updates a database
+      except transaction.interfaces.TransientError:
+          attempts += 1
+          if attempts == max_attempts:
+              raise
+      else:
+          break
+
+In the example above, we used ``transaction.manager`` to refer to the
+thread-local transaction manager, which we then used used with the
+``with`` statement.  When a conflict error occurs, the transaction
+must be aborted before retrying the update. Using the transaction
+manager as a context manager in the ``with`` statement takes care of this
+for us.
+
+The example above is rather tedious.  There are a number of tools to
+automate transaction retry.  The `transaction
+<http://zodb.readthedocs.io/en/latest/transactions.html#retrying-transactions>`_
+package provides a context-manager-based mechanism for retrying
+transactions::
+
+  for attempt in transaction.manager.attempts():
+      with attempt:
+          ... code that updates a database
+
+Which is shorter and simpler [#but-obscure]_.
+
+For Python web frameworks, there are WSGI [#wtf-wsgi]_ middle-ware
+components, such as `repoze.tm2
+<https://pypi.python.org/pypi/repoze.tm2>`_ that align transaction
+boundaries with HTTP requests and retry transactions when there are
+transient errors.
+
+For applications like queue workers or `cron jobs
+<https://en.wikipedia.org/wiki/Cron>`_, conflicts can sometimes be
+allowed to fail, letting other queue workers or subsequent cron-job
+runs retry the work,
 
 Conflict resolution
 ~~~~~~~~~~~~~~~~~~~
 
 ZODB provides a conflict-resolution framework for merging conflicting
-changes.  Commonly used objects that implement conflict resolution are
+changes.  When conflicts occur, conflict resolution is used, when
+possible, to resolve the conflicts without raising a ConflictError to
+the application.
+
+Commonly used objects that implement conflict resolution are
 buckets and ``Length`` objects provided by the `BTree
 <https://pythonhosted.org/BTrees/>`_ package.
 
 The main data structures provided by BTrees: BTrees and TreeSets,
 spread their data over multiple objects.  The leaf-level objects,
-called *buckets* allow distinct keys to be updated without causing
+called *buckets*, allow distinct keys to be updated without causing
 conflicts [#usually-avoids-conflicts]_.
 
 ``Length`` objects are conflict-free counters, that merge changes by
@@ -376,6 +430,12 @@ Some things to keep in mind when utilizing multiple processes:
   <http://relstorage.readthedocs.io/en/latest/>`_, or `NEO
   <http://www.neoppod.org/>`_, that supports multiple processes.  None
   of the included storages do.
+
+.. [#but-obscure] But also a bit obscure.  The Python context-manager
+   mechanism isn't a great fit for the transaction-retry use case.
+
+.. [#wtf-wsgi] `Web Server Gateway Interface
+   <http://wsgi.readthedocs.io/en/latest/>`_
 
 .. [#usually-avoids-conflicts] Conflicts can still occur when buckets
    split due to added objects causing them to exceed their maximum size.
--- a/doc/guide/transactions-and-threading.rst
+++ b/doc/guide/transactions-and-threading.rst
@@ -166,10 +166,10 @@ block as a transaction::
 
    >>> exec(src)
 
-This opens a connection, assignes it it's own context manager, and
+This opens a connection, assignes it its own context manager, and
 executes the nested code in a transaction.  We used ``as conn2`` to
-get the connection.  The transaction boundaries are defined ``with``
-statement.
+get the connection.  The transaction boundaries are defined by the
+``with`` statement.
 
 Getting a connection's transaction manager
 ------------------------------------------
@@ -272,7 +272,7 @@ something like this::
 
   max_attempts = 3
   attempts = 0
-  while 1:
+  while True:
       try:
           with transaction.manager:
               ... code that updates a database
@@ -311,7 +311,7 @@ transient errors.
 For applications like queue workers or `cron jobs
 <https://en.wikipedia.org/wiki/Cron>`_, conflicts can sometimes be
 allowed to fail, letting other queue workers or subsequent cron-job
-runs retry the work,
+runs retry the work.
 
 Conflict resolution
 ~~~~~~~~~~~~~~~~~~~
--- a/doc/guide/transactions-and-threading.rst
+++ b/doc/guide/transactions-and-threading.rst
@@ -6,7 +6,7 @@ Transactions and concurrency
 
 `Transactions <https://en.wikipedia.org/wiki/Database_transaction>`_
 are a core feature of ZODB.  Much has been written about transactions,
-and we won't go into much detail here.  Transactions provide 2 core
+and we won't go into much detail here.  Transactions provide two core
 benefits:
 
 Atomicity
@@ -21,7 +21,7 @@ Concurrency
   Transactions provide a way of managing concurrent updates to data.
   Different programs operate on the data independently, without having
   to use low-level techniques to moderate their access. Coordination
-  and synchronization happens via transactions.
+  and synchronization happen via transactions.
 
 
 .. _using-transactions-label:
@@ -51,7 +51,7 @@ If we decide we don't want to commit a transaction, we can use
 ``abort``::
 
   conn.root.x = 2
-  transaction.abort()
+  transaction.abort() # conn.root.x goes back to 1
 
 .. -> src
 
@@ -68,8 +68,8 @@ explanation.  When using transactions, there are three kinds of
 objects involved:
 
 Transaction
-   Transactions represent units of work.  They have beginnings and
-   ends. They provide the
+   Transactions represent units of work.  Each transaction has a beginning and
+   an end. Transaction provide the
    :interface:`~transaction.interfaces.ITransaction` interface.
 
 Transaction manager
@@ -83,7 +83,7 @@ Transaction manager
 Data manager
    Data managers manage data associated with transactions.  ZODB
    connections are data managers.  The details of how they interact
-   with transactions isn't important here.
+   with transactions aren't important here.
 
 Explicit transaction managers
 -----------------------------
@@ -91,7 +91,7 @@ Explicit transaction managers
 ZODB connections have transaction managers associated with them when
 they're opened. When we call the database :meth:`~ZODB.DB.open` method
 without an argument, a thread-local transaction manager is used. Each
-thread has it's own transaction manager.  When we called
+thread has its own transaction manager.  When we called
 ``transaction.commit()`` above we were calling commit on the
 thread-local transaction manager.
 
@@ -99,15 +99,16 @@ Because we used a thread-local transaction manager, all of the work in
 the transaction needs to happen in the same thread.  Similarly, only
 one transaction can be active in a thread.
 
-If we want to have transactions who's work is spread over multiple
-threads, or if we wanted to run multiple simultaneous transactions in
-a single thread, then we can create transaction managers ourselves and
-pass them to :meth:`~ZODB.DB.open`::
+If we want to run multiple simultaneous transactions in a single
+thread, or if we want to spread the work of a transaction over
+multiple threads [#bad-idea-using-multiple-threads-per-transaction]_,
+then we can create transaction managers ourselves and pass them to
+:meth:`~ZODB.DB.open`::
 
-  tm = transaction.TransactionManager()
-  conn = db.open(tm)
+  my_transaction_manager = transaction.TransactionManager()
+  conn = db.open(my_transaction_manager)
   conn.root.x = 2
-  tm.commit()
+  my_transaction_manager.commit()
 
 .. -> src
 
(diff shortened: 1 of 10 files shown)


jimfulton:
Note that when this is merged, we will have achieved parity with (and more accuracy than) the old guide.

jamadden:
Refs #108. 

I will review when I get a chance.

Maybe merge this first then I'll come back with a separate PR for gevent?

jimfulton:
On Sep 10, 2016 5:06 PM, "Jason Madden" <notifications@github.com> wrote:
>
> Refs #108.
>
> I will review when I get a chance.
>
> Maybe merge this first then I'll come back with a separate PR for gevent?

Yup.

Jim

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
+============================
+Transactions and concurrency
+============================
+
+.. contents::
+
+`Transactions <https://en.wikipedia.org/wiki/Database_transaction>`_
+are a core feature of ZODB.  Much has been written about transactions,
+and we won't go into much detail here.  Transactions provide 2 core
```
jamadden: 2 -> two

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+  application using the transactional system doesn't have to undo
+  partial changes.  This takes a significant burden from developers
+  and increases the reliability of applications.
+
+Concurrency
+  Transactions provide a way of managing concurrent updates to data.
+  Different programs operate on the data independently, without having
+  to use low-level techniques to moderate their access. Coordination
+  and synchronization happens via transactions.
+
```
jamadden: Seeing the *A* of ACID, I was expecting to see the *C*, but instead of *consistent* we get *concurrency*...and no *I* or *D*. Makes my head hurt :)

I know, it's simplified and the *D* is rolled into the *A* and the *I* into the *C*.

Pretty good simplification IMO.

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+  transaction to fail, the updates are rolled back automatically. The
+  application using the transactional system doesn't have to undo
+  partial changes.  This takes a significant burden from developers
+  and increases the reliability of applications.
+
+Concurrency
+  Transactions provide a way of managing concurrent updates to data.
+  Different programs operate on the data independently, without having
+  to use low-level techniques to moderate their access. Coordination
+  and synchronization happens via transactions.
```
jamadden: happens -> happen 

(A and B happen; B happens)

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+
+.. -> src
+
+   >>> exec(src)
+   >>> conn.root.x
+   1
+   >>> conn.close()
+
+In this example, because we aborted the transaction, the value of
+``conn.root.x`` was rolled back to 1.
```
jamadden: I think that deserves to be demonstrated to the user, not just hidden away in the doctest comment. That seems like a key point.

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+
+There are a number of things going on here that deserve some
+explanation.  When using transactions, there are three kinds of
+objects involved:
+
+Transaction
+   Transactions represent units of work.  They have beginnings and
+   ends. They provide the
```
jamadden: "Each transaction has a beginning and an end."?

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+   provide APIs to start and end transactions.  The transactions
+   managed are always sequential. There is always exactly one active
+   transaction associated with a transaction manager at any point in
+   time. Transaction managers provide the
+   :interface:`~transaction.interfaces.ITransactionManager` interface.
+
+Data manager
+   Data managers manage data associated with transactions.  ZODB
+   connections are data managers.  The details of how they interact
+   with transactions isn't important here.
```
jamadden: isn't -> aren't.

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+
+Explicit transaction managers
+-----------------------------
+
+ZODB connections have transaction managers associated with them when
+they're opened. When we call the database :meth:`~ZODB.DB.open` method
+without an argument, a thread-local transaction manager is used. Each
+thread has it's own transaction manager.  When we called
```
jamadden: its

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+``transaction.commit()`` above we were calling commit on the
+thread-local transaction manager.
+
+Because we used a thread-local transaction manager, all of the work in
+the transaction needs to happen in the same thread.  Similarly, only
+one transaction can be active in a thread.
+
+If we want to have transactions who's work is spread over multiple
```
jamadden: who's -> whose

Although I would phrase it as: If we want to spread the work of a transaction over multiple threads,...

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+  tm.commit()
+
+.. -> src
+
+   >>> exec(src)
+
+In this example, to commit our work, we called ``commit()`` on the
+transaction manager we created and passed to :meth:`~ZODB.DB.open`.
+
+context managers
```
jamadden: Title case.

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+----------------
+
+In the examples above, the transaction beginnings were
+implicit. Transaction's were effectively
```
jamadden: Transaction's -> Transactions

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+   >>> exec(src)
+
+A more modern [#context-managers-are-new]_ way to manage transaction
+boundaries is to use context managers and the Python ``with``
+statement. Transaction managers are context managers, so we can use
+them with the ``with`` statement directly::
+
+  with tm as trans:
+     trans.note("incrementing x")
+     conn.root.x += 1
```
jamadden: Ok, that was confusing to me on first read.

We previously associated the connection with the `tm` transaction manager, and now we're explicitly managing a transaction for the transaction manager that the connection is associated with. Lots of hidden associations there.

To be honest, we pretty much always associate the life of a transaction and a connection directly. We don't keep a connection open across multiple transaction boundaries. It's just easier to treat them as the same thing.  Our "middleware" essentially always does:
```python
transaction.begin()
with db.open() as conn:
    do_stuff()
transaction.commit()
```

In the early days, getting "cannot close a connection joined to a transaction" was a frequent, confusing error, so we simplified the rules and just manage them as one.

Maybe you'll get to that. I haven't finished reading the docs yet...I'm trying to give my reactions on a first pass through to emulate a typical new reader.

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+
+.. -> src
+
+   >>> exec(src)
+   >>> db.history(conn.root()._p_oid)[0]['description']
+   'incrementing x again'
+
+Here, we used the
+:meth:`~transaction.interfaces.ITransactionManager.get` method to get
+the current transaction.
```
jamadden: I know what you wrote about using this attribute being preferred. We never do use this attribute, we exclusively use thread/greenlet local transaction/managers. But it makes sense. (Does that attribute become None when the connection is closed? I've never checked. I assume it does.)

We do join other resource managers to them when we don't have access to a connection (and wouldn't want to). I suppose that's what you meant about "distributed transactions".

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+  >>> conn.root.x
+  3
+
+This is because it's still in the same transaction that was implicitly
+begun when a change was last committed against it.  If we want to see
+changes, we have to begin a new transaction:
+
+  >>> trans = tm.begin()
+  >>> conn.root.x
+  5
```
jamadden: This is probably part of why we always tie the lifetime of a connection and transaction together. This implicit behaviour is non-obvious until you see it documented.

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+*optimistic* approach, because it works best if there are no
+conflicts.
+
+The best way to avoid conflicts is to design your application so that
+multiple connections don't update the same object at the same time.
+This isn't always easy.
+
+Sometimes you may need to queue some operations that update shared
+data structures, like indexes, so the updates can be made by a
+dedicated thread or process.
```
jamadden: "can be made by a dedicated thread or process *sequentially, one at a time*." To be explicit about why you queue. Often task queues do things in parallel these days.

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+
+Conflict resolution
+~~~~~~~~~~~~~~~~~~~
+
+ZODB provides a conflict-resolution framework for merging conflicting
+changes. This is implemented by `BTree
```
jamadden: That almost makes it sound like they're the *only* objects that implement this. Given your disclaimer below that might be your goal, but if not it could be made more palatable by combining this paragraph with the last paragraph:

Two very commonly used objects that implement conflict resolution are BTrees and Length objects.

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+<https://pythonhosted.org/BTrees/>`_ buckets and ``Length`` objects.
+
+The main data structures provided by BTrees: BTrees and TreeSets,
+spread their data over multiple objects.  The leaf-level objects,
+called *buckets* allow distinct keys to be updated without causing
+conflicts [#usually-avoids-conflicts]_.
+
+``Length`` objects are conflict key counters, that merge changes by
+simply accumulating changes.
```
jamadden: Shout-out to https://github.com/zopefoundation/zope.minmax ?
zc.queue also resolves conflicts, maybe that deserves a pointer here and/or in the section above about queueing tasks. (zc.catalogqueue does too, but that's probably too specialized to mention...well, maybe not, you did mention indexing.)

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+
+Concurrency, threads and processes
+==================================
+
+ZODB supports concurrency through transactions.  Multiple programs can
+operate independently in separate transactions.  They synchronize at
+transaction boundaries.
+
+The most common way to run ZODB is with each program running in it's
+own thread.  Usually using the thread-local transaction manager is used.
```
jamadden: 'using' seems unnecessary.

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+and use your own transaction manager, as described in `Explicit
+transaction managers`_.  To run multiple transaction managers
+simultaneously in a thread, you need to use a separate transaction
+manager for each transaction.
+
+To spread a transaction over multiple threads, you need to keep in
+mind that database connections, transaction managers and transactions
+are **not thread-safe**.  You have to prevent simultaneous access from
+multiple threads.  For this reason, **using multiple threads with a
+single connection is not recommended**, but it is possible with care.
```
jamadden: Hard to over-emphasize that point. It might even deserve a call-ahead in the "Explicit transaction manager" section.

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
...
+rolled back.
+
+A secondary benefit of savepoints is that they save any changes made
+before the savepoint to a file, so that memory of changed objects can
+be freed if they aren't used later in the transaction.
+
+Concurrency, threads and processes
+==================================
+
+ZODB supports concurrency through transactions.  Multiple programs can
```
jamadden: programs -> tasks?

programs is an overloaded word, many people might think of operating system processes.

doc/reference/transaction.rst
```diff
@@ -0,0 +1,25 @@
+============
+Transactions
+============
+
+Transaction support is provided by the `transaction
+<http://transaction.readthedocs.io/en/latest/>`_ package, which is
+installed automatically when you install ZODB.  There are 2 important
```
jamadden: 2 -> two

doc/guide/transactions-and-threading.rst
```diff
@@ -0,0 +1,394 @@
+============================
+Transactions and concurrency
+============================
+
+.. contents::
+
+`Transactions <https://en.wikipedia.org/wiki/Database_transaction>`_
+are a core feature of ZODB.  Much has been written about transactions,
```
jamadden: A footnote or callout that the transaction package is separate and can be (and is) used by other things? It's a nice package.

jamadden:
Looks generally pretty good. I left a lot of comments, some about typos, some about things I found confusing; I wonder if we can recommend simplest best practices up front and save the rest of it for later.

Of course we might disagree on what simplest best practices are 😄

doc/guide/transactions-and-threading.rst
jamadden: A paragraph about the *other* way to deal with conflicts, through retry, would probably be good. It's explicitly supported by the transaction package (although last time I checked that API had issues and wouldn't work for us) and numerous generic packages (repoze.tm2, nti.transaction) and framework-specific packages (pyramid_tm).

doc/guide/transactions-and-threading.rst
jamadden: It seems that attribute [doesn't](https://github.com/zopefoundation/ZODB/blob/67388dcd9bab50051510e72c76b8ed29a8773a9c/src/ZODB/Connection.py#L299) get nulled-out when the connection is closed. Maybe it should be. 

In the case of lots of threads, or more often, new greenlets, that would save a bit of memory if the thread-local TM can then go away.

doc/guide/transactions-and-threading.rst
jimfulton: That's why I spell it out on words.  I'll add a comment on the abort line as well.

doc/guide/transactions-and-threading.rst
jimfulton: I like your wording.

doc/guide/transactions-and-threading.rst
jimfulton: The code would probably be easier to understand if I used ``my_transaction_manager`` rather than ``tm``. 

There are a number of issues at play here, confused a bit by ZODB's history and complicated by the ``with`` statements inadequacy for this use case. :)

In the future, I want to add an API to run functions as transaction with retry. https://groups.google.com/forum/#!topic/python-transaction/3eUzBzX2sRU
I also want to make the result of ``db.transaction()`` callable with a function (useable as a decorator) to run a function as a transaction with a connection.  But until these exist, I wanted to document what exists.

Perhaps it would be better not to mention using transaction managers as context managers.

I *should* mention that a number of frameworks provide tools for managing connections and transactions.

P.S. nti.transaction needs more documentation. :)

doc/guide/transactions-and-threading.rst
jimfulton: WRT thread/greenlet local (globals :), I understand and endorse their use pragmatically in many cases, however, I don't like using globals and try to find ways to avoid them where practical.  Code can use ``transaction_manager`` without making any assumptions about how an application is managing transactions.

jimfulton: WRT clearing the connection transaction manager, +1, less from a resource-management perspective than from a way to maybe avoid opportunities to fail, especially when a TM was a (local) global.

(It's funny to refer to a global as thread/greenlet local, but it makes sense because it's a thread/greenlet local global. :) )

doc/guide/transactions-and-threading.rst
jimfulton: Well, the real issue here isn't that it's implicit, it's really that the transaction start time determines the view of the database.

doc/guide/transactions-and-threading.rst
jimfulton: WRT dedicated thread doing things "sequentially, one at a time", the thread doesn't have to do things one at a time. In fact a major benefit of queuing updates is to match them.  I'm inclined to add ", without simultaneous updates".

jimfulton: +1 WRT discussing discussing retry logic.

doc/guide/transactions-and-threading.rst
jamadden: The fact that a transaction is automatically immediately begun (and thus implicitly freezes a connection's view of the database) when the previous transaction is committed is the non-obvious (implicit 😄) thing, I think.

Explicit is better than implicit, after all; it's why SQL has `BEGIN TRANSACTION`, and even though autocommit is a thing, why it's not usually recommended.

doc/guide/transactions-and-threading.rst
jimfulton: s/connection/transaction/ :)

I don't know what you mean by a call-ahead, but I can add a warning there.

doc/guide/transactions-and-threading.rst
jamadden: > In fact a major benefit of queuing updates is to match them. I'm inclined to add ", without simultaneous updates"

Ahh, very good point. You're right.

doc/guide/transactions-and-threading.rst
jamadden: > Perhaps it would be better not to mention using transaction managers as context managers.

Hmm, tough call. It is one of the safer ways to use them, at least in theory, although it's hardly complete on its own.

> I should mention that a number of frameworks provide tools for managing connections and transactions.

Maybe if you do that, the section about context managers can be minimized to "here's an example, but you probably want to use one of these instead."

doc/guide/transactions-and-threading.rst
jimfulton: This is the CS use of the word. Tasks is similarly overloaded. <shrug>

I'll add a footnote to try clarify.

doc/guide/transactions-and-threading.rst
jamadden: > Code can use transaction_manager without making any assumptions about how an application is managing transactions.

True enough. Just as non-ZODB code can use transactions through the thread local/global API without being aware of ZODB connections, so long as they're also using the local/global manager. But if you start to mix the two, well, Here There Be Dragons.

I tend to assume code is not ZODB aware, so the global/local API it is.

doc/guide/transactions-and-threading.rst
jimfulton: Well, we aren't referring to the package here.

Did anything else ever use the transaction package?

jamadden: [zope.sqlalchemy](https://github.com/zopefoundation/zope.sqlalchemy) is one I'm immediately aware of. There's also `repoze.sendmail`.

jimfulton: I'll add a footnote in the transaction reference.

doc/guide/transactions-and-threading.rst
jimfulton: The fact that the view is frozen has nothing to do with the transaction beginning implicitly. An explicit begin wouldn't have changed anything.

Auto-commit is an entirely different beast.

I've rarely seen BEGIN TRANSACTION used in SQL code I've read (regardless of autocommit, which again is a different beast).

doc/guide/transactions-and-threading.rst
jimfulton: I need to ponder this.  Part of the problem is that I'm not happy with the state of the art, which makes it a bit hard to document.

doc/guide/transactions-and-threading.rst
```diff
@@ -68,8 +68,8 @@ explanation.  When using transactions, there are three kinds of
 objects involved:
 
 Transaction
-   Transactions represent units of work.  They have beginnings and
-   ends. They provide the
+   Transactions represent units of work.  Each transaction has a beginning and
+   an end. Transaction provide the
```
jamadden: "Transactions provide" or "Transaction provides"

doc/guide/transactions-and-threading.rst
jimfulton: I ended up keeping the description of using transaction managers as context managers, because it made it a bit easier to provide a retry example.  I also used a longer variable name for the transaction manager, which hopefully helps.

doc/guide/transactions-and-threading.rst
jimfulton: I added a section on retry logic.

I even added an ``attempts`` example, even though I find ``attempts`` to be a bit of an incantation. :)

jimfulton:
Thank you very much for the review!

doc/guide/transactions-and-threading.rst
jamadden: I like the section.

PR #117, opened 2016-09-12
Noteworthy non-included storages: split base storages from others, NEO

jmuchemb:
In this PR, there's still a missing commit to add a entry for NEO. BTW, shall we sort alphabetically :)

### DIFF
--- a/doc/reference/storages.rst
+++ b/doc/reference/storages.rst
@@ -144,24 +144,31 @@ Options:
 Noteworthy non-included storages
 ================================
 
-A number of important ZODB storages are distributed separately, including:
+A number of important ZODB storages are distributed separately.
+
+Base storages
+-------------
+
+Unlike the included storages, all the implementations listed in this section
+allow multiple processes to share the same database.
 
 RelStorage
   `RelStorage <http://relstorage.readthedocs.io/en/latest/>`_
   stores data in relational databases.  This is especially
   useful when you have requirements or existing infrastructure for
-  storing data in relational databases.  Unlike the included storages,
-  multiple processes can share the same database.
+  storing data in relational databases.
 
   For more information, see http://relstorage.readthedocs.io/en/latest/.
 
 ZEO
   `ZEO <https://github.com/zopefoundation/ZEO>`_ is a client-server
   database implementation for ZODB.  To use ZEO, you run a ZEO server,
-  and use ZEO clients in your application.  Unlike the included
-  storages, multiple processes can share the same database.
+  and use ZEO clients in your application.
 
   For more information, see https://github.com/zopefoundation/ZEO.
+
+Optional layers
+---------------
 
 ZRS
   `ZRS <https://github.com/zc/zrs>`_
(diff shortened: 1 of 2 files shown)


doc/reference/storages.rst
```diff
@@ -135,7 +135,7 @@ Demo storages are configured using the ``demostorage`` section::
 ``demostorage`` sections can contain up to 2 storage subsections,
 named ``base`` and ``changes``, specifying the demo storage's base and
 changes storages.  See :meth:`ZODB.DemoStorage.DemoStorage.__init__`
-for more on the base anc changes storages.
+for more on the base and changes storages.
```
jmuchemb: ... for more *information* on ... ?

jimfulton: See ZODB.DemoStorage.DemoStorage.__init__
for more on the base and changes storages.

jimfulton:
I'm fine with sorting alphabetically. :)

jmuchemb:
For me, it can be merged.

Maybe all those "For more information, see <url>" are too much. We could either remove them (because each first occurrence of the storage name points to the same url), or replace by

    homepage: <url>

doc/reference/storages.rst
```diff
@@ -144,31 +144,46 @@ Options:
...
+-------------
+
+Unlike the included storages, all the implementations listed in this section
+allow multiple processes to share the same database.
+
+NEO
+  `NEO <https://lab.nexedi.com/nexedi/neoppod>`_ can spread data among several
+  computers for load-balancing and multi-master replication. It also supports
+  asynchronous replication to off-site NEO databases for further disaster
+  resistance without affecting local operation latency.
```
jamadden: I would consider it important to mention that of all the storages mentioned here, NEO is the only one that's Linux only; all the rest of them are operating-system independent. (The linux-only requirement is the major reason my organization can't test NEO---we develop on OS X.)

jmuchemb: I'm not sure it's the place to start listing drawbacks of each solution. The main limitations (like linux-only but also no blob support) look visible enough on the linked page.

The use of `epoll()` is the only reason why it requires Linux. In 2 places:
  https://lab.nexedi.com/nexedi/neoppod/blob/master/neo/lib/event.py
  https://lab.nexedi.com/nexedi/neoppod/blob/master/neo/tests/threaded/__init__.py

On our side, it's the contrary: we only develop and run test suites on Linux. We're open to contributions but unless someone provides test bots for other OS, we could only maintain it if the alternative implementation is not OS-specific, like select or libevent.

I'll see with @vpelletier

jamadden: I wasn't trying to criticize that as a drawback, just highlight that it's a substantial difference between that storage and the others. But then again, so is blob support.

You are right, it is quite clear from the linked page. So I withdraw my suggestion (unless someone else still feels like it needs to be included).

jimfulton: I'm fine with the NEO text as is.

jimfulton:
I assume this is complete now. LGTM.

PR #118, opened 2016-09-14
Add a section on the pitfalls of __eq__/__hash__. Fixes #106.

### LINKED ISSUE #106: Documentation: Writing persistent objects: Be careful about __eq__ and __hash__
In general, objects without custom `__eq__` and `__hash__` objects are going to be friendlier on the DB and the cache (if you can get away with identity semantics).

Suppose you have two classes:
```python
import uuid

class WithHash(Persistent):
   def __init__(self):
      self.id = uuid.uuid()
   def __eq__(self, other):
      return self.id == other.id
   def __hash__(self):
      return hash(self.id)

class WithoutHash(Persistent):
    def __init__(self):
       self.id = uuid.uuid()
```

Now if you had some dictionaries using those classes as keys:
```python
conn = db.open()
conn.root.with_hash_dict = {WithHash(): i for i in range(5000)}
conn.root.wout_hash_dict = {WithoutHash(): i for i in range(5000)}

```

Doing something like:
```python
conn = db.open()
len(conn.root.with_hash_dict)
```
is going to unghost 5000 `WithHash` objects, whereas

```python
conn = db.open()
len(conn.root.wout_hash_dict)
```
isn't going to unghost *any* objects (because unpickling a dictionary re-hashes all the keys, and  `hash()` looks up `__hash__` on the class not the instance, so if the `__hash__` method doesn't access any attributes---like the default one---nothing has to be unghosted) .

Even if your object cache is sized appropriately, large-ish dictionaries can take a long time to unpickle when accessed for the first time in a particular connection, adding lots of load to the DB and/or cache system; the same happens when creating a dictionary in memory for the first time of such persistent objects.

If you can accept identity semantics (and for persistent objects, you surprisingly often can), it's better to avoid custom `__eq__` and `__hash__` methods if you'll ever be creating dictionaries or sets of your persistent objects.

This is a lesson we learned the hard way; coming from a Java background almost all of our objects defined custom `__eq__` and `__hash__` methods, and that was fine until we started to get a lot of objects, when it became a performance burden. Now it turns out that many such of those objects don't need these methods.

Worth adding to the docs?

jamadden:
Here's a first draft. Comments welcome!

### DIFF
--- a/doc/guide/writing-persistent-objects.rst
+++ b/doc/guide/writing-persistent-objects.rst
@@ -574,24 +574,22 @@ For example, a class that defines ``__eq__`` and ``__hash__`` like this::
 
 is going to be much slower to use as a key in a persistent dictionary,
 or in a new dictionary when the key is a ghost, than the class that
-inherits identity-based ``__eq__`` and ``__hash__``::
+inherits identity-based ``__eq__`` and ``__hash__``.
 
-  class Book(persistent.Persistent):
+.. Example of the above.
 
-     def __init__(self, title):
-         self.title = title
-         self.authors = ()
+    Here's what that class would look like::
 
-     def add_author(self, author):
-         self.authors += (author, )
-
-.. -> src
-
-    >>> exec(src)
+    >>> class Book(persistent.Persistent):
+    ...    def __init__(self, title):
+    ...        self.title = title
+    ...        self.authors = ()
+    ...
+    ...    def add_author(self, author):
+    ...        self.authors += (author, )
 
     Lets see an example of how these classes behave when stored in a
     dictionary. First, lets store some dictionaries::
-
 
     >>> import ZODB
     >>> db = ZODB.DB(None)
--- a/doc/guide/writing-persistent-objects.rst
+++ b/doc/guide/writing-persistent-objects.rst
@@ -1,6 +1,6 @@
-==========================
-Writing persistent objects
-==========================
+============================
+ Writing persistent objects
+============================
 
 In the :ref:`Tutorial <tutorial-label>`, we discussed the basics of
 implementing persistent objects by subclassing
@@ -11,7 +11,7 @@ of writing persistent classes you should be aware of.
 Access and modification
 =======================
 
-Two of the main jobs of the ``Persistent`` base class is to detect
+Two of the main jobs of the ``Persistent`` base class are to detect
 when an object has been accessed and when it has been modified.  When
 an object is accessed, its state may need to be loaded from the
 database.  When an object is modified, the modification needs to be
@@ -214,6 +214,15 @@ Here's a version of the example that uses a ``TreeSet``::
    >>> bool(book.authors._p_changed)
    True
    >>> db.close()
+
+If you're going to use custom classes as keys in a ``BTree`` or
+entries in a ``TreeSet``, they must provide a `total ordering
+<https://pythonhosted.org/BTrees/#total-ordering-and-persistence>`_.
+The builtin python `str` class is always safe to use as BTree key. You
+can use `zope.keyreference
+<https://pypi.python.org/pypi/zope.keyreference>`_ to treat arbitrary
+persistent objects as totally orderable based on their persistent
+object identity.
 
 Scalable sequences are a bit more challenging. The `zc.blist
 <https://pypi.python.org/pypi/zc.blist/>`_ package provides a scalable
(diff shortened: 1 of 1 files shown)


doc/guide/writing-persistent-objects.rst
```diff
@@ -503,6 +512,121 @@ ghost:
...
+When that ``dict`` is later loaded from the database, the internal
+storage is rebuild from scratch. This means that every key has its
+``__hash__`` method called at least once, and may have its ``__eq__``
+method called many times.
+
+By default, every object, including persistent objects, inherits an
+implementation of ``__eq__`` and ``__hash__`` from :class:`object`.
+These default implementations are based on the object's *identity*,
+that is, its unique identifier within the current Python process.
+Calling, them, therefore is very fast, even on ghosts, and doesn't
```
jimfulton: make one of the uses a "ghosts" uses a reference to the section on ghosts, using:

    :ref:`ghost <ghost-label>`

doc/guide/writing-persistent-objects.rst
```diff
@@ -503,6 +512,121 @@ ghost:
...
+
+While you can do anything with a persistent subclass that you can with
+a normal subclass, certain things have additional implications for
+persistent objects. These often show up as performance issues.
+
+Implement ``__eq__`` and ``__hash__``
+-------------------------------------
+
+When you store an entry into a Python ``dict`` (or the persistent
+variant ``PersistentMapping``) the key's ``__eq__`` and
```
jimfulton: or set

doc/guide/writing-persistent-objects.rst
```diff
@@ -503,6 +512,121 @@ ghost:
...
+  class Book(persistent.Persistent):
+
+     def __init__(self, title):
+         self.title = title
+         self.authors = ()
+
+     def add_author(self, author):
+         self.authors += (author, )
+
+Lets see an example of how these classes behave when stored in a
```
jimfulton: This example is tl;dr. The reader doesn't need to see the proof.  Move the example into the comment above (after exec(src).)

Same goes for the class above.

I used to think that maybe doctests made could documentation. They don't. :)

I do like making docs testable, but the proofs should be hidden in comments.

doc/guide/writing-persistent-objects.rst
```diff
@@ -503,6 +512,121 @@ ghost:
...
+    >>> all((book._p_status == 'ghost' for book in conn2.root.with_ident))
+    True
+    >>> all((book._p_status == 'ghost' for book in conn2.root.with_hashes))
+    False
+
+
+We can see that all the objects that did have a custom ``__eq__``
+and ``__hash__`` were loaded into memory, while those that did weren't.
+
+There are two possible solutions:
```
jimfulton: Maybe:

Some alternatives:

- Ask whether you really need to persistent objects as keys.

...

doc/guide/writing-persistent-objects.rst
```diff
@@ -503,6 +512,121 @@ ghost:
...
+
```
jimfulton: To me, the compelling case for persistent objects as keys is **really** using them in sets.

It just occurred to me that this is related to the discussion on the list WRT oid generation.  It would be nice to use oids and index ids, *especially* if oids were allocated randomly.

jimfulton:
Thanks for writing this.  It's a bit long, but will be helped by moving the proofs into comments. :)

doc/guide/writing-persistent-objects.rst
```diff
@@ -503,6 +512,121 @@ ghost:
...
+By default, every object, including persistent objects, inherits an
+implementation of ``__eq__`` and ``__hash__`` from :class:`object`.
+These default implementations are based on the object's *identity*,
+that is, its unique identifier within the current Python process.
+Calling, them, therefore is very fast, even on ghosts, and doesn't
+cause a ghost to load its state.
+
+If you override ``__eq__`` and ``__hash__`` in a custom persistent
+subclass, however, when you use that instances of that class as a key
+in a ``dict``, then the instance will have to be a unghosted before it
```
jzuech3: to be unghosted

doc/guide/writing-persistent-objects.rst
```diff
@@ -503,6 +512,121 @@ ghost:
...
+
+By default, every object, including persistent objects, inherits an
+implementation of ``__eq__`` and ``__hash__`` from :class:`object`.
+These default implementations are based on the object's *identity*,
+that is, its unique identifier within the current Python process.
+Calling, them, therefore is very fast, even on ghosts, and doesn't
+cause a ghost to load its state.
+
+If you override ``__eq__`` and ``__hash__`` in a custom persistent
+subclass, however, when you use that instances of that class as a key
```
jzuech3: `use that instances` -> use an instance

doc/guide/writing-persistent-objects.rst
```diff
@@ -503,6 +512,121 @@ ghost:
     >>> book._p_changed, bool(book._p_oid)
     (None, True)
 
+Things you can do, but need to carefully consider (advanced)
+============================================================
+
+While you can do anything with a persistent subclass that you can with
+a normal subclass, certain things have additional implications for
+persistent objects. These often show up as performance issues.
```
jmuchemb: Maybe the 2 advanced sections could be merged:
```diff
diff --git a/doc/guide/writing-persistent-objects.rst b/doc/guide/writing-persistent-objects.rst
index 568b17f..0e7fa0a 100644
--- a/doc/guide/writing-persistent-objects.rst
+++ b/doc/guide/writing-persistent-objects.rst
@@ -517,7 +517,8 @@ Things you can do, but need to carefully consider (advanced)
 
 While you can do anything with a persistent subclass that you can with
 a normal subclass, certain things have additional implications for
-persistent objects. These often show up as performance issues.
+persistent objects. These often show up as performance issues, or the
+resulting may become hard to maintain.
 
 Implement ``__eq__`` and ``__hash__``
 -------------------------------------
@@ -626,13 +627,6 @@ There are two possible solutions:
   nature of a BTree means that only a small number of objects will
   have to be loaded in most cases.
 
-
-Other things you can do, but shouldn't (advanced)
-=================================================
-
-The first rule here is don't be clever!!!  It's very tempting to be
-clever, but it's almost never worth it.
-
 Overriding ``__getstate__`` and ``__setstate__``
 ------------------------------------------------
 

```

jamadden:
Thanks for the feedback, I've made changes to address it.

doc/guide/writing-persistent-objects.rst
```diff
@@ -583,8 +585,12 @@ inherits identity-based ``__eq__`` and ``__hash__``::
      def add_author(self, author):
          self.authors += (author, )
 
-Lets see an example of how these classes behave when stored in a
-dictionary. First, lets store some dictionaries::
+.. -> src
```
jimfulton: End the previous sentence with a period and move the class into the proof. The reader doesn't need to see the class that *doesn't* define __eq__ and __hash__. :)

jamadden: Got it.

jimfulton:
Thanks for writing this!

PR #120, opened 2016-09-24
Enable using coverage and coveralls for coverage reporting.

jamadden:
Set this up on travis.

This was just slightly tricky because of the use of 'j99' in the call to bin/test.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,8 +1,9 @@
 [run]
-# source = ...
-# omit = ...
-# cover_pylib = False
-# branch = True
+source = src/ZODB/
+parallel = true
+omit =
+	 src/ZODB/tests/*
+	 src/ZODB/scripts/tests/*
 
 [report]
 exclude_lines =
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ dist
 testing.log
 .eggs/
 .dir-locals.el
+htmlcov
+tmp
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,13 +16,22 @@ matrix:
         - os: linux
           python: 3.5
 install:
+    - pip install -U pip
     - pip install zc.buildout
     - buildout $BUILOUT_OPTIONS
 script:
-    - bin/test -v1j99
-    - if [[ $TRAVIS_PYTHON_VERSION != 'pypy3' ]]; then cd doc; make html; fi
+    - if [[ $TRAVIS_PYTHON_VERSION != pypy* ]]; then bin/coverage run bin/coverage-test -v1j99; fi
+    - if [[ $TRAVIS_PYTHON_VERSION == 'pypy' || $TRAVIS_PYTHON_VERSION == 'pypy3' ]]; then bin/test -v1j99; fi
+    - if [[ $TRAVIS_PYTHON_VERSION != 'pypy3' ]]; then pushd doc; make html; popd; fi
+    - if [[ $TRAVIS_PYTHON_VERSION != pypy* ]]; then pip install coveralls; fi # install early enough to get into the cache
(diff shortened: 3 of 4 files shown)


jamadden:
Travis is green and coveralls is reporting! Mostly it's the 'scripts' that are dragging the coverage numbers down, but there are a few other questionable things; this might be a tool to help remove unneeded code (I notice with sadness that RelStorage is still using `UndoLogCompatible`).

jamadden:
Any thoughts @jimfulton ?

jimfulton:
Thanks!

tseaver:
Should we be adding a Coveralls button to the README?  Looks like the other buttons are brokey, too.

jamadden:
My experience with the buttons is that they're fine and good in the repo, but when they start to wind up on PyPI (which happens when the PyPI page is the same as the repo's README.rst), it becomes One More Thing that you have to change every time you do a new release, or at least a new branch (e.g., 4.x vs 5.x). Otherwise you wind up with badges for 4.1.2 pointing to the build and coverage status for master, and if master is ever broken for any reason your PyPI page looks bad. Doing it on a branch-by-branch basis is *almost* doable (I try with gevent) but pinning it to a particular release is annoyingly painful, and without that you're still vulnerable to some unexpected breakage making your page look bad. 

The contrary is also possible with Coveralls. The coverage for master might be much higher than the coverage for an older branch, leading to a false sense of confidence.

gevent may be a particularly painful example for this because it has some very timing sensitive tests that can break if the machine is too slow. But IIRC there have been times that ZODB or ZEO tests on Travis have appeared broken through no direct fault of a code change for similar reasons.

All the development information just moved to 'DEVELOPERS.rst' from README.rst, and thus off PyPI. Maybe the buttons could go there? Of course, then they're no longer displayed to visitors to the repo by default...

Or maybe the PyPI info should be completely separate from the repo info. The audiences are probably different.

mgedmin:
On Tue, Sep 27, 2016 at 06:52:41PM -0700, Jason Madden wrote:
> My experience with the buttons is that they're fine and good in the repo, but
> when they start to wind up on PyPI (which happens when the PyPI page is the
> same as the repo's README.rst), it becomes One More Thing that you have to
> change every time you do a new release, or at least a new branch (e.g., 4.x vs
> 5.x). Otherwise you wind up with badges for 4.1.2 pointing to the build and
> coverage status for master, and if master is ever broken for any reason your
> PyPI page looks bad. Doing it on a branch-by-branch basis is almost doable (I
> try with gevent) but pinning it to a particular release is annoyingly painful,
> and without that you're still vulnerable to some unexpected breakage making
> your page look bad.

TBH as a developer, whenever I see Travis buttons on PyPI I think they
represent the state of master, not the state of the last release.

(Also I think it's a good sign for the package to be using Travis.  Even
if the badge is currently red.)

> The contrary is also possible with Coveralls. The coverage for master might be
> much higher than the coverage for an older branch, leading to a false sense of
> confidence.
> 
> gevent may be a particularly painful example for this because it has some very
> timing sensitive tests that can break if the machine is too slow. But IIRC
> there have been times that ZODB or ZEO tests on Travis have appeared broken
> through no direct fault of a code change for similar reasons.
> 
> All the development information just moved to 'DEVELOPERS.rst' from README.rst,
> and thus off PyPI. Maybe the buttons could go there? Of course, then they're no
> longer displayed to visitors to the repo by default...

You could put the buttons in README.rst and then have setup.py strip
them off with a regexp or something.

Marius Gedminas
-- 
Professionalism has no place in art, and hacking is art.  Software Engineering
might be science; but that's not what I do.  I'm a hacker, not an engineer.
        -- jwz

PR #121, opened 2016-09-27
Simplify the README file to avoid out of date information

jimfulton:
We have the same information scattered around in different places,
which increases the chance that some of it will be out of date.

See the end of:
https://groups.google.com/forum/#!topic/zodb/fy6RRVAF9-s

I want to try to avoid duplicating zodb.org.


jamadden:
+1. LGTM. I'm all for simpler PyPI pages that point to more easily updated docs.

Developers.rst should be updated at some point to mention virtualenvs and the semi-deprecation of bootstrap.py, but that can be a separate PR. It'd also be nice if that information made it to the website too, eventually (he says, suddenly realizing that gevent's own developer docs are only in the README/on PyPI).

PR #123, opened 2016-10-18
Clean DB properties on close

agroszer:
otherwise persistent object seem to hang around
noticed by our tests

### DIFF
--- a/src/ZODB/DB.py
+++ b/src/ZODB/DB.py
@@ -646,7 +646,7 @@ class DB(object):
         del self.storage
         del self._mvcc_storage
         # clean up references to other DBs
-        self.databases.clear()
+        self.databases = {}
         # clean up the connection pool
         while self.pool.pop():
             pass
--- a/src/ZODB/DB.py
+++ b/src/ZODB/DB.py
@@ -107,6 +107,10 @@ class AbstractConnectionPool(object):
 
     size = property(getSize, lambda self, v: self.setSize(v))
 
+    def clear(self):
+        pass
+
+
 class ConnectionPool(AbstractConnectionPool):
 
     def __init__(self, size, timeout=1<<31):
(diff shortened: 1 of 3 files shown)


src/ZODB/DB.py
```diff
@@ -645,6 +645,15 @@ def _(c):
         self._mvcc_storage.close()
         del self.storage
         del self._mvcc_storage
+        # clean up references to other DBs
+        self.databases.clear()
+        # clean up the connection pool
+        while self.pool.pop():
+            pass
```
tseaver: Maybe `AbstractConnectionPool` should grow a `clear()` method:  we would then just call it for each one:

```python
        self.pool.clear()
        self.historical_pool.clear()
```

src/ZODB/tests/multidb.txt
```diff
@@ -137,7 +137,7 @@ an exception:
 
 Clean up:
 
-    >>> for a_db in dbmap.values():
+    >>> for a_db in list(dbmap.values()):
```
tseaver: Nice find.

src/ZODB/DB.py
```diff
@@ -645,6 +645,15 @@ def _(c):
         self._mvcc_storage.close()
         del self.storage
         del self._mvcc_storage
+        # clean up references to other DBs
+        self.databases.clear()
```
jamadden: I wonder if this should be `self.databases = {}` or even `self.databases = None`? That way if someone else has a reference to the databases dict (not hard if you're manually constructing a multi-db) it doesn't get cleared out from under them.

jimfulton: +1

jimfulton: In fact, I just looked at some code that clearing the dict would break.

@agroszer just reset the attr as @jamadden suggested.

strichter: Done.

src/ZODB/DB.py
strichter: Done. It was easy enough.

strichter:
@jamadden & @tseaver Are you okay with my implementation of your comments? I would love to see this getting merged and a release being made. :smile:

jimfulton:
Gaaaa, we didn't check for an entry in CHANGES.rst. I just added one.

jamadden:
Snap. Sorry I missed that.

jimfulton:
Released in 5.2.4.

PR #127, opened 2016-11-13
Cleaned up check_checkCurrentSerialInTransaction

jimfulton:
- Use assertEqual rather than assertTrue for better error messages.

- Fixed a typo that caused an assertion to be missed.


jimfulton:
Oh, coverage was reduced! By changing a few characters that made an assertion actually assert something useful. sigh

jimfulton:
An effect from the transaction package change, which I guess missed before, (spuriously) broke a test. Fixed.

PR #128, opened 2016-11-16
zodbcmp - Tool to compare two ZODB databases

navytux:
This is a tool to compare two ZODB databases in between tidmin..tidmax
transaction range with default range being -∞..+∞ - (whole database).

For comparision both databases are scanned at storage layer and every
transaction content is compared bit-to-bit between the two. The program stops
either at first difference found, or when whole requested transaction range is
scanned with no difference detected.

Database storages are specified in files with ZConfig-based storage definition, e.g.

    %import neo.client
    <NEOStorage>
        master_nodes    ...
        name            ...
    </NEOStorage>

Please see https://lab.nexedi.com/nexedi/neoppod/merge_requests/4 for
one of possible contexts.

The tool is generic though and is not NEO-specific. It should be able to
even check two different storages like ZEO & NEO, or FileStorage and NEO
etc and thus can be handy.

### DIFF
--- a/src/ZODB/scripts/zodbcmp.py
+++ b/src/ZODB/scripts/zodbcmp.py
@@ -1,0 +1,217 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016  Nexedi SA and Contributors.
+#                     Kirill Smelkov <kirr@nexedi.com>
+"""Zodbcmp - Tool to compare two ZODB databases
+
+Zodbcmp compares two ZODB databases in between tidmin..tidmax transaction range
+with default range being -∞..+∞ - (whole database).
+
+For comparison both databases are scanned at storage layer and every
+transaction content is compared bit-to-bit between the two. The program stops
+either at first difference found, or when whole requested transaction range is
+scanned with no difference detected.
+
+Exit status is 0 if inputs are the same, 1 if different, 2 if error.
+"""
+
+from __future__ import print_function
+from time import time
+
+def ashex(s):
+    return s.encode('hex')
+
+# something that is greater than everything else
+class Inf:
+    def __cmp__(self, other):
+        return +1
+inf = Inf()
+
+
+# get next item from iter -> (item, !stop)
+def nextitem(it):
+    try:
+        item = it.next()
+    except StopIteration:
+        return None, False
+    else:
+        return item, True
+
+# objects of a IStorageTransactionInformation
+def txnobjv(txn):
+    objv = []
+    for obj in txn:
+        assert obj.tid == txn.tid
+        assert obj.version == ''
+        objv.append(obj)
+
+    objv.sort(key = lambda obj: obj.oid)    # in canonical order
+    return objv
+
+
+# compare two storage transactions
+# 0 - equal, 1 - non-equal
+def txncmp(txn1, txn2):
+    # metadata
+    for attr in ('tid', 'status', 'user', 'description', 'extension'):
+        attr1 = getattr(txn1, attr)
+        attr2 = getattr(txn2, attr)
+        if attr1 != attr2:
+            return 1
+
+    # data
+    objv1 = txnobjv(txn1)
+    objv2 = txnobjv(txn2)
+    if len(objv1) != len(objv2):
+        return 1
+
+    for obj1, obj2 in zip(objv1, objv2):
+        for attr in ('oid', 'data', 'data_txn'):
+            attr1 = getattr(obj1, attr)
+            attr2 = getattr(obj2, attr)
+            if attr1 != attr2:
+                return 1
+
+    return 0
+
+
+# compare two storages
+# 0 - equal, 1 - non-equal
+def storcmp(stor1, stor2, tidmin, tidmax, verbose=False):
+    iter1 = stor1.iterator(tidmin, tidmax)
+    iter2 = stor2.iterator(tidmin, tidmax)
+
+    Tprev = time()
+    txncount = 0
+    while 1:
+        txn1, ok1 = nextitem(iter1)
+        txn2, ok2 = nextitem(iter2)
+
+        # comparison finished
+        if not ok1 and not ok2:
+            if verbose:
+                print("equal")
+            return 0
+
+        # one part has entry not present in another part
+        if txn1 is None or txn2 is None or txn1.tid != txn2.tid:
+            if verbose:
+                tid1 = txn1.tid if txn1 else inf
+                tid2 = txn2.tid if txn2 else inf
+                l = [(tid1, 1,2), (tid2, 2,1)]
+                l.sort()
+                mintid, minstor, maxstor = l[0]
+                print("not-equal: tid %s present in stor%i but not in stor%i" % (
+                        ashex(mintid), minstor, maxstor))
+            return 1
+
+        # show current comparison state and speed
+        if verbose:
+            txncount += 1
+            T = time()
+            if T - Tprev > 5:
+                print("@ %s  (%.2f TPS)" % (ashex(txn1.tid), txncount / (T - Tprev)))
+                Tprev = T
+                txncount = 0
+
+        # actual txn comparison
+        tcmp = txncmp(txn1, txn2)
+        if tcmp:
+            if verbose:
+                print("not-equal: transaction %s is different")
+            return 1
+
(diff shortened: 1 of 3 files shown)


src/ZODB/scripts/zodbcmp.py
```diff
@@ -0,0 +1,213 @@
...
+        if txn1 is None or txn2 is None or txn1.tid != txn2.tid:
+            if verbose:
+                tid1 = txn1.tid if txn1 else inf
+                tid2 = txn2.tid if txn2 else inf
+                l = [(tid1, 1,2), (tid2, 2,1)]
+                l.sort()
+                mintid, minstor, maxstor = l[0]
+                print("not-equal: tid %s present in stor%i but not in stor%i" % (
+                        ashex(mintid), minstor, maxstor))
+            return 1
```
jmuchemb:
```python
        txn1 = next(iter1, None)
        txn2 = next(iter2, None)
        if txn1 != txn2:
            if verbose:
                mintid, minstor, maxstor = min(
                    (txn1.tid if txn1 else inf, 1, 2),
                    (txn2.tid if txn2 else inf, 2, 1))
                print(...)
            return 1
        if not txn1:
            if verbose:
                print("equal")
            return 0
```
`MAX_TID` (instead of `inf`) may also be enough.

src/ZODB/scripts/zodbcmp.py
```diff
@@ -0,0 +1,213 @@
...
+            txncount += 1
+            T = time()
+            if T - Tprev > 5:
+                print("@ %s  (%.2f TPS)" % (ashex(txn1.tid), txncount / (T - Tprev)))
+                Tprev = T
+                txncount = 0
+
+        # actual txn comparison
+        tcmp = txncmp(txn1, txn2)
+        if tcmp:
```
jmuchemb: I really wonder why you like temp variables so much.

src/ZODB/scripts/zodbcmp.py
```diff
@@ -0,0 +1,213 @@
...
+    tidmax = tidmax.decode('hex')
+
+    # empty tid means -inf / +inf respectively
+    # ( which is None in IStorage.iterator() )
+    if not tidmin:
+        tidmin = None
+    if not tidmax:
+        tidmax = None
+
+    return tidmin, tidmax
```
jmuchemb:
```python
    return (tidmin.decode('hex') or None,
            tidmax.decode('hex') or None)
```

src/ZODB/scripts/zodbcmp.py
```diff
@@ -0,0 +1,213 @@
...
+
+def main():
+    try:
+        main2()
+    except:
+        exc_type, _, _ = sys.exc_info()
+        if exc_type is SystemExit:
+            raise   # this was sys.exit() call, not an error
+        traceback.print_exc()
+        sys.exit(2)
```
jmuchemb:
```python
    try:
        main2()
    except SystemExit:
        raise
    except:
        traceback.print_exc()
        sys.exit(2)
```

navytux: ok, this was a thinko on my side, thanks for correction.

src/ZODB/scripts/zodbcmp.py
navytux: Thanks for the `or None` hint. However since .split("..") and .decode("hex") raise different exceptions, for proper error reporting I decided to wrap both & hex-decode actions and reaise dedicated exception so users will get e.g.

```
E: invalid tidrange: q..
```

instead of

```
  File "./zodbcmp.py", line 205, in <module>
    main()
  File "./zodbcmp.py", line 192, in main
    tidmin, tidmax = parse_tidrange(argv[2])
  File "./zodbcmp.py", line 153, in parse_tidrange
    tidmin = tidmin.decode('hex')
  File ".../lib/python2.7/encodings/hex_codec.py", line 42, in hex_decode
    output = binascii.a2b_hex(input)
TypeError: Odd-length string
```

See the change in interdiff for next patch revision.

src/ZODB/scripts/zodbcmp.py
navytux: I've thought about MAX_TID too but:

- there is no MAX_TID in ZODB (it is in NEO) - in ZODB it is `utils.maxtid`
- in both ZODB and NEO it is 7fff... instead of ffff... for signedness. Still having 7f vs ff looks there can be a risk of skipping transactions.
- I wanted code here to be genric, not depending on particular MAX value a tid can be. This can be achieved with `inf` approach.

src/ZODB/scripts/zodbcmp.py
navytux: It makes the code more structured and clear to me.

navytux:
@jmuchemb thanks for feedback. I've adjusted the patch. The interdiff is below:

```diff
diff --git a/src/ZODB/scripts/zodbcmp.py b/src/ZODB/scripts/zodbcmp.py
index f70739e73..bc16a6e24 100755
--- a/src/ZODB/scripts/zodbcmp.py
+++ b/src/ZODB/scripts/zodbcmp.py
@@ -147,19 +147,24 @@ def usage(out):
 """, file=out)

 # tidmin..tidmax -> (tidmin, tidmax)
+class TidRangeInvalid(Exception):
+    pass
+
 def parse_tidrange(tidrange):
-    tidmin, tidmax = tidrange.split("..")
-    tidmin = tidmin.decode('hex')
-    tidmax = tidmax.decode('hex')
+    try:
+        tidmin, tidmax = tidrange.split("..")
+    except ValueError:  # not exactly 2 parts in between ".."
+        raise TidRangeInvalid(tidrange)
+
+    try:
+        tidmin = tidmin.decode("hex")
+        tidmax = tidmax.decode("hex")
+    except TypeError:   # hex decoding error
+        raise TidRangeInvalid(tidrange)

     # empty tid means -inf / +inf respectively
     # ( which is None in IStorage.iterator() )
-    if not tidmin:
-        tidmin = None
-    if not tidmax:
-        tidmax = None
-
-    return tidmin, tidmax
+    return (tidmin or None, tidmax or None)

 def main2():
     verbose = False
@@ -189,8 +194,8 @@ def main2():
     if len(argv) > 2:
         try:
             tidmin, tidmax = parse_tidrange(argv[2])
-        except ValueError:
-            usage(sys.stderr)
+        except TidRangeInvalid as e:
+            print("E: invalid tidrange: %s" % e, file=sys.stderr)
             sys.exit(2)

     stor1 = ZODB.config.storageFromFile(open(storconf1, 'r'))
@@ -202,10 +207,9 @@ def main2():
 def main():
     try:
         main2()
+    except SystemExit:
+        raise   # this was sys.exit() call, not an error
     except:
-        exc_type, _, _ = sys.exc_info()
-        if exc_type is SystemExit:
-            raise   # this was sys.exit() call, not an error
         traceback.print_exc()
         sys.exit(2)
```

jimfulton:
This sounds like a useful tool, but I'd rather it not be included in ZODB.  This has nothing to do with the tool, which I haven't studied :), but with the weight of ZODB maintenance.  This weight has been historically been onerous in the case of utility scripts, which like this one, have lacked tests and tended to rot.

If it were up to me, the scripts directory would go away.

FWIW, I hope that soon there will be some sort of namespace package for the ZODB ecosystem.  This could be a tool to release separately within that namespace. (Although I would prefer it have tests.)

navytux:
@jimfulton thanks for feedback. It could eventually have tests like checking database against itself is equal and aginst another db a bit modified - as not equal. It's a pity you do not want core (imho) ZODB-related things to be in ZODB repository itself. Just for the reference `scripts/fstest.py` and `scripts/repozo.py` _have_ tests and `zodbcmp` could have too.

> If it were up to me, the scripts directory would go away.

but because of this I assume retrying the pull-request with tests for zodbcmp which are run on every `python setup.py test` is not the way to go. Or am I maybe wrong?

jimfulton:
At a high level, it would be better to release this separately.  At the same time, we should create a page on zodb.org (via a ZODB/docs pr :)) that makes it easy for people to find tools like this.

navytux:
ok I'm moving this to https://lab.nexedi.com/kirr/zodbtools . I hesitate to change ZODB/docs about the tool for now.

Thanks once again for feedback,
Kirill

PR #129, opened 2016-11-17
transaction user and description: text above, bytes below

jimfulton:
ZODB now translates transaction meta data, ``user`` and
``description`` from text to bytes before passing them to storages,
and converts them back to text when retrieving them from storages in
the ``history``, ``undoLog`` and ``undoInfo`` methods.

The ``IDatabase`` interface was updated to reflect that ``history``,
``undoLog`` and ``undoInfo`` are available on database objects.
(They were always available, but not documented in the interface.)


jimfulton:
Note that the majority of the changes are trivial test changes or interface changes.

jmuchemb:
I would simplify `TransactionMetaData` by doing all necessary convertion in `__init__`, instead of using `property` (only `_extension` would be a `property`). In fact, I don't understand the use of `property` for `extension`.

What about deprecating `_extension` ?

src/ZODB/DB.py
```diff
@@ -1064,3 +1064,12 @@ def connection(*args, **kw):
     managing a separate database object.
     """
     return DB(*args, **kw).open_then_close_db_when_connection_closes()
+
+transaction_meta_data_text_variables = 'user_name', 'description'
+def text_transaction_info(info):
```
jamadden: Underscore prefix these as private implementation details?

src/ZODB/interfaces.py
```diff
@@ -445,7 +558,7 @@ class IStorage(Interface):
     implement multi-version concurrency control.
 
     There are methods in IStorage and in derived interfaces that
-    provide information about the current revisions (tids) for objects
+    provide information babout the current revisions (tids) for objects
```
jamadden: typo.

src/ZODB/interfaces.py
```diff
@@ -685,6 +799,8 @@ def tpc_abort(transaction):
     def tpc_begin(transaction):
         """Begin the two-phase commit process.
 
+        The argument provides IStorageTransactionMetaDataExtensionBytes.
```
jamadden: No such interface, so far as I can see, just `IStorageTransactionMetaData`.

jimfulton:
Update: I'm simplifying ``TransactionMetaData`` as @jmuchemb suggested. I had some reasons for the properties, but they weren't good ones.

I'm also changing ``BaseStorage`` and ``MappingStorage`` to use ``extension`` rather than ``_extension``. This has led to lots of new test failures that I'm working my way through. Some if these were tests I'd failed to update and some were due to a bug in undo handling, which hadn't been updated to use TransactionMetaData.

BTW, if I'd used ``extension_info`` in ``TransactionMetaData`` (or ``extension`` in ``Transaction``), I wouldn't have caught these bugs. :)

jimfulton:
Given how backed up Travis has been :(, I'm not going to push anything until I think I'm done.

src/ZODB/DB.py
jimfulton: sure

src/ZODB/interfaces.py
jimfulton: Thanks

src/ZODB/interfaces.py
jimfulton: Yeah, fixing

src/ZODB/Connection.py
```diff
@@ -1270,3 +1285,52 @@ def __repr__(self):
...
+
+    @property
+    def extension(self):
+        return self.__extension
+
+    @extension.setter
+    def extension(self, v):
+        self.__extension = v
+
+    _extension = extension
```
jimfulton: I'm not following your comment on copyTransactionsFrom.  BaseStorage's implementation, which uses copy, passes the iterator's transaction record to tpc_begin, which copies extension data.

jimfulton:
I pushed what I have now. I'm going to be afk for a while.

I'm willing to rename ``extended_info`` to ``extension`` in ``Transaction``, but I find it perversely  useful to leave them inconsistent. :)

jamadden:
> I'm willing to rename extended_info to extension in Transaction, but I find it perversely useful to leave them inconsistent. :)

I agree that it's useful. It's also technical debt that can be confusing and makes it difficult to navigate (at least, I got myself turned around 😄 ). I'm willing to do the PRs against transaction and this package as needed (presumably once this PR lands for this package; I think the PR against transaction will apply cleanly to master with or without zopefoundation/transaction#32).

jimfulton:
I'll go ahead and make a transaction pr.

jimfulton:
Done

jimfulton:
Are we good now?

jamadden:
I took another look at everything, and it all looks good to me!...

...With one tiny quibble (sorry). IIUC, ZODB.Connection.TransactionMetaData is the implementation of IStorageTransactionMetaData. If so, could it get that @implementer declaration? (I find that helps make navigating an unfamiliar code base easier.)

src/ZODB/Connection.py
```diff
@@ -1286,6 +1287,7 @@ def __repr__(self):
 size.
 """
 
+@implementer(IStorageTransactionMetaData)
```
jamadden: Thanks!

jamadden:
Are there plans to release this soon? With transaction 2.0.3 being out, anyone that upgrades will find it's impossible to use ZODB 5.0.x:
```
  File "/home/travis/virtualenv/pypy-5.4.1/site-packages/ZODB/DB.py", line 472, in __init__
    with self.transaction('initial database creation') as conn:
  File "/home/travis/virtualenv/pypy-5.4.1/site-packages/ZODB/DB.py", line 1015, in __enter__
    t.note(self.note)
  File "/home/travis/virtualenv/pypy-5.4.1/site-packages/transaction/_transaction.py", line 537, in note
    raise TypeError("Note must be text (unicode)")
TypeError: Note must be text (unicode)
```

jimfulton:
Done

PR #132, opened 2016-11-18
Fixed: ZODB.Connection.TransactionMetaData didn't custom data

jimfulton:
storage that some storages rely on.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,9 @@
  Change History
 ================
 
+- Fixed: ``ZODB.Connection.TransactionMetaData`` didn't custom data
+  storage that some storages rely on.
+
 5.1.0 (2016-11-17)
 ==================
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,7 +2,7 @@
  Change History
 ================
 
-- Fixed: ``ZODB.Connection.TransactionMetaData`` didn't custom data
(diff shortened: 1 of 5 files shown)


CHANGES.rst
```diff
@@ -2,6 +2,9 @@
  Change History
 ================
 
+- Fixed: ``ZODB.Connection.TransactionMetaData`` didn't custom data
```
jamadden: Missing some words?

src/ZODB/Connection.py
```diff
@@ -1321,3 +1321,22 @@ def _extension(self):
...
+    def data(self, ob):
+        try:
+            data = self._data
+        except AttributeError:
+            raise KeyError(ob)
+
+        try:
+            return data[id(ob)]
+        except KeyError:
+            raise KeyError(ob)
```
jamadden: Is this simpler and equivalent (fewer lines of code to cover)?
```python
try:
   return self._data[id(ob)]
except (AttributeError, KeyError):
   raise KeyError(ob)
```

Not that it's a big deal. If you prefer the more explicit version that's fine by me.

jimfulton: Sure. This code was copied verbatim from transaction.

CHANGES.rst
jimfulton: yup

PR #135, opened 2016-12-20
Allow creating ghosts with data from search results

jimfulton:
The connection get method now accept an optional class pickle as  a second argument.  This is a somewhat experimental feature to support index and search of objects in external indexes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,10 @@
  Change History
 ================
 
+- The connection ``get`` method now accept an optional class pickle as
+  a second argument.  This is a somewhat experimental feature to
+  support index and search of objects in external indexes.
+
 5.1.1 (2016-11-18)
 ==================
 
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -227,7 +227,7 @@ class Connection(ExportImport, object):
         # registered with the transaction.
         self._added[oid] = obj
 
-    def get(self, oid):
+    def get(self, oid, class_pickle=None):
         """Return the persistent object with oid 'oid'."""
         if self.opened is None:
             raise ConnectionStateError("The database connection is closed")
@@ -242,7 +242,10 @@ class Connection(ExportImport, object):
         if obj is not None:
             return obj
 
-        p, _ = self._storage.load(oid)
+        if class_pickle is None:
+            p, _ = self._storage.load(oid)
+        else:
+            p = class_pickle
         obj = self._reader.getGhost(p)
 
         # Avoid infiniate loop if obj tries to load its state before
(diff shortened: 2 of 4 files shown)


jmuchemb:
I wonder what's your use case to prefer a class pickle as a second argument, instead of the class itself. I admit that only the class is not enough when there are __new__ args but so far, pickles have remained hidden to the upper layer (the application using the ZODB).

jamadden:
If this is useful functionality, I would like to see it as a completely separate method instead of adding more responsibilities on to `get`. Internally the two functions might share an implementation, but I don't think that the keyword parameter approach that makes `get` completely bypass the storage makes for a good public API.

I've seen this discussion come up on the python mailing lists before, and I like GVRs philosophy: if any given call site is only going to be calling the function in one way, then instead of an optional argument you probably want two functions. 

A separate function (maybe even documented in a separate interface, `IDangerousConnectionStuff`) could emphasize the relative footgun-ness of this functionality, instead of making the normal, safe `get` function sometimes do exotic things.

jimfulton:
On Wed, Dec 21, 2016 at 4:52 AM, Julien Muchembled <notifications@github.com
> wrote:

> I wonder what's your use case to prefer a class pickle as a second
> argument, instead of the class itself. I admit that only the class is not
> enough when there are *new* args but so far,
>
Exactly.

> pickles have remained hidden to the upper layer (the application using the
> ZODB).
>
I don't expect applications to use this.  This is at somewhat the same
level as storages, but addressing different needs.

jimfulton:
On Wed, Dec 21, 2016 at 7:47 AM, Jason Madden <notifications@github.com>
wrote:

> If this is useful functionality, I would like to see it as a completely
> separate method instead of adding more responsibilities on to get.
> Internally the two functions might share an implementation, but I don't
> think that the keyword parameter approach that makes get completely
> bypass the storage makes for a good public API.
>
get is barely a public API. I can't remember the last time I wrote
application code that used it. <shrug>

> I've seen this discussion come up on the python mailing lists before, and
> I like GVRs philosophy: if any given call site is only going to be calling
> the function in one way, then instead of an optional argument you probably
> want two functions.
>
He doesn't stick to that as often as you might think, but in any case,
whether something does something substantially different is often a
subjective decision.

A separate function (maybe even documented in a separate interface,
> IDangerousConnectionStuff) could emphasize the relative footgun-ness of
> this functionality, instead of making the normal, safe get function
> sometimes do exotic things.
>
I'm fine with this.  Given that I think the use case is somewhat
experimental (but promising), If I made this a separate method, I'd not
bother making it part of a contract for now.

I'll do this.

jimfulton:
On 3rd though, I think I'll withdraw this and monkey-patch initially.

PR #136, opened 2017-01-13
Fix deadlock with storages that "sync" on a new transaction

jmuchemb:
This backports a change from commit 227953b977a9e195c4ce9bbb9acd9c5ee60c333a.

NEO, as well as ZEO+server_sync (ERP5 backports this feature with a monkey-patch), pings the server (primary master node in the case of NEO) on new transactions. However, this round-trip is actually performed by the thread that also does tasks requiring to lock the DB, like processing of invalidations.

Since transaction 1.6.1 (more precisely zopefoundation/transaction@e581a120a612d54eea9240606a5723ff73952eb0), IStorage.sync() is called indirectly by DB.open() when a transaction has already begun, and the DB must not be locked when this happens.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,10 +7,14 @@
 
 - ``persistent`` is no longer required at setup time.
   See `issue 119 <https://github.com/zopefoundation/ZODB/issues/119>`_.
+
 - ``Connection.close`` and ``Connection.open`` no longer race on
   ``self.transaction_manager``, which could lead to
   ``AttributeError``. This was a bug introduced in 5.0.1. See `issue
   142 <https://github.com/zopefoundation/ZODB/pull/143>`_.
+
+- Fix deadlock with storages that "sync" on a new transaction.
+  https://github.com/zopefoundation/ZODB/pull/136
 
 4.4.4 (2016-11-27)
 ==================
--- a/src/ZODB/DB.py
+++ b/src/ZODB/DB.py
@@ -753,18 +753,15 @@ class DB(object):
                     result = self.pool.pop()
             assert result is not None
 
-            # open the connection.
-            result.open(transaction_manager)
-
             # A good time to do some cache cleanup.
             # (note we already have the lock)
             self.pool.availableGC()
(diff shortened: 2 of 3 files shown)


jmuchemb:
For the failure on pypy3, see pypa/setuptools#864 :(

jamadden:
In .travis.yml, to get a Python 3.3 compatible version of pypy3,  `pypy3` can instead be spelled `pypy3.3-5.2-alpha1`. We've done that in some other projects (recently in BTrees). (I suspect, but haven't tested, that `pypy3.3-5.5-alpha` will also work for a newer version.)

jmuchemb:
Travis is green with pypy3.3-5.2-alpha1

But how to know the list of available python ? It's really not nice that they provide alpha versions instead of final ones.

The master branch does not have issue with pypy3 because it builds in a different way (buildout).

jamadden:
> But how to know the list of available python ?

I don't have a definitive answer.

I do know that the default mapping (e.g., "2.7 -> 2.7.12") is defined in [this file](https://github.com/travis-ci/travis-cookbooks/blob/master/cookbooks/travis_python/attributes/default.rb). (There's a lag between that file and the defaults going into effect.)

[This file](https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/python.rb) shows that anything not pre-installed will be downloaded from a URL that begins with https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/12.04/x86_64/. Unfortunately you can't just browse that URL to see available versions I think, but you can make educated guesses and probe to see what's up there.

pyenv is also pre-installed, it turns out, so you can use any one of the versions listed in [here](https://github.com/yyuu/pyenv/tree/master/plugins/python-build/share/python-build), with a little extra work.

jmuchemb:
I backported def73970b7873a4460719b4466e97e935e55bc9a & 5dcca55e2834d570f66b59f792578cc112187f27 (use buildout to build so we don't have 2 environments to manage) from the 5.x branch, so the pypy3 issue gone without asking Travis to use an alpha version.

PR #137, opened 2017-01-21
Define the official expansion of "ZODB"

jimfulton:
I want to tell people to stop calling it the Zope Object Database.
(Maybe we'll rename it someday, who knows, but in the mean time...)
But then I realized that the docs are silent on this, so I decided to
fix this.


doc/index.rst
```diff
@@ -201,7 +201,18 @@ Learning more
...
+* `The ZODB Book (in progress) <http://zodb.readthedocs.org/en/latest/>`_
+
+What is the expansion of "ZODB"?
+================================
+
+The expansion of "ZODB" is the Z Object Database.  But, of course, we
+usually just use "ZODB".
+
+In the past, it was the Zope Object Database, because it was
+developed at part of the Zope project.  But ZODB doesn't depend on
```
jamadden: at -> as

doc/index.rst
```diff
@@ -201,7 +201,18 @@ Learning more
...
+Zope in any way and is used my many projects that have nothing to do
```
jamadden: my -> in

jimfulton:
Haha, coverage increased because of a change in the docs.

jamadden:
I suspect the coverage thing is due to rounding and the use of too many digits. Notice the next build decreased again.

As for the change, I guess I'm at about a -0. There's a lot of existing material (e.g., https://en.wikipedia.org/wiki/ZODB --- which redirects to https://en.wikipedia.org/wiki/Zope_Object_Database) that would need to change to begin to make this effective, and it raises questions about other projects (ZEO *does* define itself as "Zope Enterprise Objects". Is that because of Zope corporation or Zope-the-platform, and if so, should it change?).

Maybe something sufficiently zen can side step the knock-on effects? Something like "ZODB does not stand for anything. In the past, people referred to it as the Zope Object Database because of its close association with the Zope application server, but ZODB does not have any specific ties to that server and can be, and is, used independently."

PR #145, opened 2017-02-07
Add a DB option to require (and leverage) explicit transaction boundaries

jimfulton:
See https://github.com/zodb/relstorage/issues/147

### DIFF
--- a/src/ZODB/DB.py
+++ b/src/ZODB/DB.py
@@ -363,7 +363,7 @@ class DB(object):
                  databases=None,
                  xrefs=True,
                  large_record_size=1<<24,
-                 explicit_transactions = False,
+                 explicit_transactions=False,
                  **storage_args):
         """Create an object database.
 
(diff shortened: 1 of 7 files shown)


src/ZODB/DB.py
```diff
@@ -363,6 +363,7 @@ def __init__(self,
                  databases=None,
                  xrefs=True,
                  large_record_size=1<<24,
+                 explicit_transactions = False,
```
freddrake: What's with the inconsistent spacing?

jimfulton: Sigh, I missed that. This is one of the places where PEP 8 (and Guildo) are wrong.

jimfulton:
This might be simplified if I implemented the proposal I made on the transaction list yesterday.

Rather than configuring this via the DB. Connection objects could get their explicitness via the transaction managers passed to them.

jimfulton:
Yeah, I think it would be better to control this via the TM.

PR #146, opened 2017-02-08
Notify storages that transactions have completed.

jimfulton:
See: https://github.com/zodb/relstorage/issues/147

- Call new afterCompletion API on storages to allow them to free
  resources after transaction complete.  See:
  https://github.com/zodb/relstorage/issues/147

- Take advantage of the new transaction-manager explicit mode to avoid
  starting transactions unnecessarily when transactions end.

This differs from the previous PR:

- No longer implement seatbelts, as the transaction package does that.

- No longer configure explicit mode via DB.  We get explicit mode from the transaction manager.

### DIFF
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -746,9 +746,6 @@ class Connection(ExportImport, object):
         except AttributeError:
             assert self._storage is None
 
-        # Now is a good time to collect some garbage.
-        self._cache.incrgc()
-
     def afterCompletion(self, transaction):
         # Note that we we call newTransaction here for 2 reasons:
         # a) Applying invalidations early frees up resources
@@ -763,6 +760,9 @@ class Connection(ExportImport, object):
 
         if not self.explicit_transactions:
             self.newTransaction(transaction, False)
+
+        # Now is a good time to collect some garbage.
+        self._cache.incrgc()
 
     # Transaction-manager synchronization -- ISynchronizer
     ##########################################################################
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -902,22 +902,25 @@ class Connection(ExportImport, object):
             # New code is in place.  Start a new cache.
             self._resetCache()
 
-        # This newTransaction is to deal with some pathalogical cases:
-        #
-        # a) Someone opens a connection when a transaction isn't
-        #    active and proceeeds without calling begin on a
-        #    transaction manager. We initialize the transaction for
-        #    the connection, but we don't do a storage sync, since
-        #    this will be done if a well-nehaved application calls
-        #    begin, and we don't want to penalize well-behaved
-        #    transactions by syncing twice, as storage syncs might be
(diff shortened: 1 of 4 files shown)


jimfulton:
There's a place where we're still calling newTransaction that we shouldn't. I'm trying to get a test to confirm the fix. :/

jimfulton:
OK, fixed.

src/ZODB/Connection.py
```diff
@@ -757,7 +758,11 @@ def afterCompletion(self, transaction):
         #    finalizing previous ones without calling begin.  We pass
         #    False to avoid possiblyt expensive sync calls to not
         #    penalize well-behaved applications that call begin.
-        self.newTransaction(transaction, False)
+        if hasattr(self._storage, 'afterCompletion'):
+            self._storage.afterCompletion()
```
jamadden: This is going to result in deferring `storage.sync`, `storage.poll_invalidations` and `cache.incrgc` for non-explicit transactions. It seems like the reasoning for point (a) should still apply though, and so maybe those things should still be done before `afterCompletion`?

[In the RelStorage case (post zopefoundation/relstorage#167) where explicit transactions *are* on, on the surface this looks like it would result in rolling back the load connection twice, but it actually won't because RelStorage handles that case already.]

jimfulton: I don't understand what you're saying.    Are you referring to the lines above? or below? (I'm guessing below.(

We're only deferring in the explicit case.

I would think that the cost of polling remote servers outweighs the benefits of maybe invalidating things.

I agree that doing a cache gc here makes sense.

I also wonder about the wisdom of doing cache gc at the *beginning* of a transaction.  Perhaps the cache gc should be moved from new to here.

jimfulton: WRT "rolling back twice", I wondered about that a little.  It's conceivable that this could do work if you started a new transaction right after ending one, but avoiding that seems like the responsibility of the storage.  Ultimately, this is like an event that the storage can do with what it will.

jamadden: > Are you referring to the lines above? or below? (I'm guessing below.)

Yeah. But I couldn't move the comment after I started it (this started out to be a different comment. :)

> We're only deferring in the explicit case.

True. Typo artifact in my comment.

> I would think that the cost of polling remote servers outweighs the benefits of maybe invalidating things.

Could be. I don't have good intuition about that.


> I agree that doing a cache gc here makes sense.

> I also wonder about the wisdom of doing cache gc at the beginning of a transaction. Perhaps the cache gc should be moved from new to here.

+1. Those both seem to make sense to me.

jamadden: > but avoiding that seems like the responsibility of the storage. Ultimately, this is like an event that the storage can do with what it will.

I agree. There's not really anything else that can be done at this level.

jimfulton:
OK, I moved incrgc. Are you good now?

Thanks for the review, as always.

Issue #148, opened 2017-02-20
Sphinx requires at least Python 2.7 or 3.4 to run.


agroszer:
That breaks buildout with py 3.3

see:
http://winbot.zope.org/builders/ZODB_dev%20py_330_win32/builds/210/steps/buildout/logs/stdio:

```
...
Installing sphinx.
Getting distribution for 'Sphinx'.
ERROR: Sphinx requires at least Python 2.7 or 3.4 to run.
error: Setup script exited with 1
An error occurred when trying to install Sphinx 1.5.2. Look above this message for any errors that were output by easy_install.
While:
  Installing sphinx.
  Getting distribution for 'Sphinx'.
Error: Couldn't install: Sphinx 1.5.2
```

icemac:
Drop support for Python 3.3 as it has its sunset in September 2017?

jimfulton:
+1 to remove 3.3, but also update the build command in travis to only build the test part.

PR #149, opened 2017-02-20
Disable Sphinx on Python 3.3

mgedmin:
See #148.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ matrix:
         - os: linux
           python: 2.7
         - os: linux
+          env: BUILOUT_OPTIONS=sphinx:eggs=
           python: 3.3
         - os: linux
           python: 3.4


jimfulton:
This is OK, but...

I think it builds on a stupid fix of mine.

To deal with this issue for pypy I copied a solution from the transaction package. (This avoided thinking, which I don't oppose on principle. :))  This fix was necessarily for transaction, IIRC, because the transaction tests use sphinx doctests.

I think a simpler fix for ZODB would be to just add ``parts=test`` to the buildout line.

mgedmin:
I believe we need two parts: `test` and `coverage-test`.

Actually I just noticed the `versions:sphinx=1.4.9` bit in `.travis.yml`, which means there's no hurry -- the next Travis build on master is not going to suddenly fail because the latest Sphinx no longer supports Python 3.3.

PR #150, opened 2017-03-09
Update RelStorage information in multi-zodb-gc document

jamadden:
RelStorage 2.0 can and should use zc.zodbdgc for reference collecting and packing in a multi-db, so long as its history free.


jimfulton:
Thanks!

PR #151, opened 2017-03-21
Add info on Newt DB

maikroeder:


jimfulton:
Oh wow, thanks. I really need to go back and revisit that.  But this is a great improvement in the meantime.

PR #154, opened 2017-04-07
Fix PyPI page

jimfulton:
I noticed the other day that the PyPI page for the current release is hosed WRT ReST:

https://pypi.python.org/pypi/ZODB/5.2.0

In fixing this, I realized that a lot of text was out of date and not well suited to a PyPI page, so I:

- Split developer info into a separate doc.

- Simplified the README and setup script.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,8 +6,8 @@
 ==================
 
 - Fixed: When opening FileStorages in read-only mode, non-existent
-  files were silently created.  Creating a file-storage against a
-  non-existent file errors.
+  files were silently created.  Creating a read-only file-storage
+  against a non-existent file errors.
 
 5.2.0 (2017-02-09)
 ==================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,14 +5,15 @@
 5.2.1 (unreleased)
 ==================
 
-- Nothing changed yet.
-
+- Fixed: When opening FileStorages in read-only mode, non-existent
+  files were silently created.  Creating a file-storage against a
+  non-existent file errors.
 
 5.2.0 (2017-02-09)
 ==================
 
 - Call new afterCompletion API on storages to allow them to free
-  resources after transaction complete.  
+  resources after transaction complete.
   See `issue 147 <https://github.com/zodb/relstorage/issues/147>`__.
 - Take advantage of the new transaction-manager explicit mode to avoid
   starting transactions unnecessarily when transactions end.
--- a/DEVELOPERS.rst
+++ b/DEVELOPERS.rst
@@ -1,0 +1,68 @@
+For developers of ZODB
+======================
+
+Building
+---------
+
+Bootstrap buildout, if necessary using ``bootstrap.py``::
+
+  python bootstrap.py
+
+Run the buildout::
+
+  bin/buildout
+
+Testing
+-------
+
+The ZODB checkouts are `buildouts <http://www.python.org/pypi/zc.buildout>`_.
+When working from a ZODB checkout, first run the bootstrap.py script
+to initialize the buildout:
+
+    % python bootstrap.py
+
+and then use the buildout script to build ZODB and gather the dependencies:
+
+    % bin/buildout
+
+This creates a test script:
+
+    % bin/test -v
+
+This command will run all the tests, printing a single dot for each
+test.  When it finishes, it will print a test summary.  The exact
+number of tests can vary depending on platform and available
+third-party libraries.::
+
+    Ran 1182 tests in 241.269s
+
+    OK
+
+The test script has many more options.  Use the ``-h`` or ``--help``
+options to see a file list of options.  The default test suite omits
+several tests that depend on third-party software or that take a long
+time to run.  To run all the available tests use the ``--all`` option.
+Running all the tests takes much longer.::
+
+    Ran 1561 tests in 1461.557s
+
+    OK
+
+Our primary development platforms are Linux and Mac OS X.  The test
+suite should pass without error on these platforms and, hopefully,
+Windows, although it can take a long time on Windows -- longer if you
+use ZoneAlarm.
+
+Generating docs
+---------------
+
+cd to the doc directory and::
+
+  make html
+
+Contributing
+------------
+
+Almost any code change should include tests.
+
+Any change that changes features should include documentation updates.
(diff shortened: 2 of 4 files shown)


CHANGES.rst
```diff
@@ -5,14 +5,15 @@
 5.2.1 (unreleased)
 ==================
 
-- Nothing changed yet.
-
+- Fixed: When opening FileStorages in read-only mode, non-existent
+  files were silently created.  Creating a file-storage against a
```
jamadden: "Creating a read-only FileStorage..."

Leaving out the 'read-only' makes that sentence sound a little scary.

setup.py
```diff
@@ -100,19 +91,11 @@ def emit(self, record):
...
-    with open(os.path.join(*file_path), 'rb') as file:
-        return file.read()
-
-long_description = str(
-    ("\n".join(doclines[2:]) + "\n\n" +
-     ".. contents::\n\n" +
-     read_file("README.rst").decode('latin-1')  + "\n\n" +
-     read_file("CHANGES.rst").decode('latin-1')))
+long_description = (
+    open("README.rst").read()  +
```
jamadden: Please bring back `read_file` so that the files get closed and not leaked on PyPy and Python 3.

setup.py
```diff
@@ -128,7 +111,7 @@ def read_file(*path):
       url = 'http://www.zodb.org/',
       license = "ZPL 2.1",
       platforms = ["any"],
-      description = doclines[0],
+      description = long_description[1],
```
jamadden:
```pycon
>>> long_description = (
...     open("README.rst").read()  +
...    "\n\n" +
...     open("CHANGES.rst").read()
...  )
>>> long_description[1]
'='
```

Doesn't seem right.

DEVELOPERS.rst
```diff
@@ -0,0 +1,68 @@
+For developers of ZODB
```
jamadden: This is no longer published anywhere on the web. Perhaps it could be referenced from a file included in doc/index.rst (.. include: ../DEVELOPERS.rst), and then it might be a good idea to add a link from the PyPI page to this info.

I'd also suggest adding an explicit link to the github repo *somewhere*, probably in this document. If I need to fix a bug (open a PR) in an unfamiliar package, I need to find its repo. First I check the PyPI page, it's often listed as the URL. But if not, I look for a "developing" page in the documentation which should list the repo. Bonus points if that just takes me one click from the PyPI page.

DEVELOPERS.rst
```diff
@@ -0,0 +1,68 @@
...
+======================
```
jamadden: Would you like to use the "standard" RST headers now that this is its own document? E.g., line above and below the document title? If you're in Emacs, C-c C-a C-s does that for you.

setup.py
jimfulton: This script is short lived. Why does that matter?

setup.py
jimfulton: oops

setup.py
jamadden: Because it's just good practice to clean up after yourself?

There are tools like pyroma and others that import the setup file and then do a bunch of analysis so the script is not necessarily short lived.

DEVELOPERS.rst
jimfulton: Well, it's published here.

+1 to adding a repo URL to the readme and making it more explicit in the docs.

Eventually, I think development should be a topic in the docs.

This PR just aims to unscrew some things.

At a meta level, If my goal was to be perfect, I would probably have put this off and the PyPI page would have remained broken longer.

DEVELOPERS.rst
jimfulton: Um, sure, whatever. :)

setup.py
jimfulton: Pyroma rns for, like, 2 seconds.

I'll add it back, but this just clutters the code. I object to cluttered code more than I object to a non-existent memory-management problem.

jamadden: That's probably an artifact of our backgrounds, then, because I object to bare `open`s and relying on the GC to close resources...curse you Java :)

jimfulton: There is no leak here.  If this script was long lived, the files would eventually be closed by the garbage collector. Even if this was long lived, which it isn't, there is only a potential problem if the program were opening files faster than the GC could reap them.

The pedantic is the enemy of the productive.

CHANGES.rst
jimfulton: fixed. Thanks.

setup.py
jimfulton: done

DEVELOPERS.rst
jimfulton: I added links to the README. I chose to make a drive-by change to the docs.

DEVELOPERS.rst
jimfulton: done

README.rst
```diff
@@ -19,3 +19,8 @@ high-degree of transparency.
 ZODB is an ACID Transactional database.
 
 To learn more, visit: http://www.zodb.org
+
+The githib repository is: at https://github.com/zopefoundation/zodb
```
jamadden: github -> github

setup.py
jimfulton: fixed

README.rst
```diff
@@ -19,3 +19,8 @@ high-degree of transparency.
...
+
+If you're interested in contributing to ZODB itself, see the
+`developer notes <DEVELOPERS.rst>`_.
```
jamadden: Although that renders just fine locally in `restview --long-description`, is that going to be a broken link on PyPI? I didn't know it would automatically render more than long-description.

setup.py
jamadden: Thank you. 

Now I'll sleep easier knowing that PyPy won't complain about this. After all, it's not like there's anything else worrying going on in the world, right? 😄

README.rst
jimfulton: Oops, fixed

README.rst
jimfulton: Gaaa, was thinking in terms of github, not pypi. Fixed.

DEVELOPERS.rst
jamadden: That works for me!

jamadden:
Thanks for this. I see that I am, and apologize for being, the one that originally introduced the typo that broke the PyPI rendering.

jimfulton:
No worries. It's fixed now.  I wonder if we can start hiding old releases now.

Issue #155, opened 2017-04-07
Why are the permissions for blobs hardcoded to give only the user access?


enfold-josh:
Currently the blobs code creates directories with permissions giving only user read, write and execute and files with permissions giving only user read. This is hardcoded and impossible to change without patching.

We have a client that wants to use solr to index content. To index blobs using collective.solr solr requires permissions to access the blobs on the filesystem. Currently without patching this would require the solr process to run as the same user as the zope processes. This is problematic because the solr process would then have write access to the blobstorage directories.

jimfulton:
This is a horrible misfeature.

jimfulton:
Note: https://github.com/zopefoundation/ZODB/pull/1

PR #156, opened 2017-04-08
Blob default permissions

jimfulton:
This basically makes blob permissions come (more or less) from whatever the user has set for new directories.

This fixes: #155 .

It replaces: #1 .

I predict windows headaches, but maybe not. We need to get appveyor working to find out. That will be a separate PR.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,9 @@
 5.2.2 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Fixed: A blob misfeature set blob permissions so that blobs and blob
+  directories were only readable by the database process owner, rather
+  than honoring user-controlled permissions (e.g. ``umask``).
 
 
 5.2.1 (2017-04-08)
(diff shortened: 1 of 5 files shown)


jimfulton:
Oh, note that I believe that this will cause a RelStorage test failure: https://github.com/zodb/relstorage/blob/master/relstorage/tests/blob/testblob.py#L434

jamadden:
Thanks for the heads-up re the RelStorage test. RelStorage's testblob.py contains parts of an oldish copy of the ZODB file, so that's not surprising. (It's not entirely clear to me how much value some of those tests have since they basically just fall through to the BlobStorageMixin or helper provided by ZODB, but I suppose they test the integration.)

As for the fix, I'll cross that bridge when we get to it 😄  I first need to resolve zodb/relstorage#169 and determine the cutoff for the oldest supported ZODB version and get tests running on it to see how to proceed (I'm sure there are installations out there on Python 2.7.8 that can't run the latest ZEO and hence ZODB, so this is of some practical importance...up until very recently I had direct knowledge of a large such installation.)

jimfulton:
Thanks!

PR #162, opened 2017-04-19
Backport from 4.0 and later

gotcha:
Avoid waiting for GC to close file. Could explain some bug with RelStorage

### DIFF
--- a/src/ZODB/blob.py
+++ b/src/ZODB/blob.py
@@ -68,7 +68,8 @@ class Blob(persistent.Persistent):
             raise TypeError('Blobs do not support subclassing.')
         self.__setstate__()
         if data is not None:
-            self.open('w').write(data)
+            with self.open('w') as f:
+                f.write(data)
 
     def __setstate__(self, state=None):
         # we use lists here because it will allow us to add and remove


jamadden:
It took substantial work to make zodb run on pypy successfully (see #34 and #17 among others), and not just in this project; changes were needed in persistent and relstorage too. A branch this old with old dependencies  is unlikely to be successful there.

gotcha:
On ZODB 3.10 with RelStorage and CPython

I see an error in `tpc_abort` when the transaction aborts after conflict.
RelStorage delegates to `shutil.rmtree` to remove the directory with a newly created blob and it stops with `OSError: [Errno 16] Device or resource busy`
I am guessing that it is because of a file left open during the transaction.

Reading the code lead me to that file not explicitely closed. My understanding is that files are closed when garbage collected. In this case, I think that I cannot be sure it will have happened when `tpc_abort` is called.

jimfulton:
On Wed, Apr 19, 2017 at 11:19 AM, Godefroid Chapelle <
notifications@github.com> wrote:

> On ZODB 3.10 with RelStorage and CPython
>
> I see an error in tpc_abort when the transaction aborts after conflict.
> RelStorage delegates to shutil.rmtree to remove the directory with a
> newly created blob and it stops with OSError: [Errno 16] Device or
> resource busy
> I am guessing that it is because of a file left open during the
> transaction.
>
> Reading the code lead me to that file not explicitely closed. My
> understanding is that files are closed when garbage collected. In this
> case, I think that I cannot be sure it will have happened when tpc_abort
> is called.
>
But in C Python, GC includes reference counting which happens immediately.
To illustrate:

>>> import ZODB
>>> db = ZODB.DB('t.fs', blob_dir='b')
>>> conn = db.open()
>>> import ZODB.blob
>>> conn.root.b = ZODB.blob.Blob()
>>> f = conn.root.b.open('w')
>>> conn.root.b.writers
[<weakref at 0x1037bc7e0; to 'BlobFile' at 0x103623bd0>]
>>> f = None
>>> conn.root.b.writers
[]

Anyway, I certainly don't object to this change.  It's good hygiene.

I'm working on getting the tests to pass again on the 3.10 branch. When
that's done, you should merge those changes into this PR and then hopefully
the tests will pass.

jimfulton:
On Wed, Apr 19, 2017 at 12:28 PM, Jim Fulton <jim@jimfulton.info> wrote:

> I'm working on getting the tests to pass again on the 3.10 branch. When
> that's done, you should merge those changes into this PR and then hopefully
> the tests will pass.
>

Well, the 3.10 tests seem to be passing in Travis, even though they fail on
my Mac. My work is done here. :)  (I spent way too much time on this.)

Jim


-- 
Jim Fulton
http://jimfulton.info

gotcha:
@jimfulton Thanks for making tests pass on Travis ! Even though the backport I propose is useless on CPython.

Any hint how I could debug the OSError I mentioned above ?

jimfulton:
Can you construct a test case/script that provokes the OSError?

PR #164, opened 2017-05-26
In py3, dict_keys type has no attribute sort

djanderson:
Using `ZODB.scripts.analyze` under python3, I ran into the follow issue:

```  File "/usr/local/lib/python3.5/dist-packages/ZODB/scripts/analyze.py", line 71, in report
    typemap.sort()
AttributeError: 'dict_keys' object has no attribute 'sort'
```

This commit uses the builtin `sorted` function instead.

### DIFF
--- a/src/ZODB/scripts/analyze.py
+++ b/src/ZODB/scripts/analyze.py
@@ -67,8 +67,7 @@ def report(rep):
     fmts = "%46s %7d %8dk %5.1f%% %7.2f" # summary format
     print(fmt % ("Class Name", "Count", "TBytes", "Pct", "AvgSize"))
     print(fmt % ('-'*46, '-'*7, '-'*9, '-'*5, '-'*7))
-    typemap = rep.TYPEMAP.keys()
-    typemap.sort()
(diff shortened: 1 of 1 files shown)


jimfulton:
The reason this broke in the first place, and the reason I hate these random scripts, is that the script was untested.

tseaver:
@jimfulton There is [at least one tests for the `anaylize` script](https://github.com/zopefoundation/ZODB/blob/master/src/ZODB/tests/testFileStorage.py#L399).

jimfulton:
On Tue, May 30, 2017 at 1:22 PM, Tres Seaver <notifications@github.com>
wrote:

> @jimfulton <https://github.com/jimfulton> There is at least one tests for
> the anaylize script
> <https://github.com/zopefoundation/ZODB/blob/master/src/ZODB/tests/testFileStorage.py#L399>
> .
>

Cool. I wonder if it could be updated to cover this case.

tseaver:
Even funnier, that one test actually [replicates the feature](https://github.com/zopefoundation/ZODB/blob/master/src/ZODB/tests/testFileStorage.py#L450) which this PR fixes.

PR #170, opened 2017-07-24
HistoricalStorageAdapter forwards release() to its instance

### LINKED ISSUE #78: HistoricalStorageAdapter does not forward release to its base
This causes a connection leak on RelStorage.

jamadden:
Fixes #78.

Also adjust the base wrapper's __getattr__ to be faster (500 vs 711ns in a benchmark of the common case) and behave the same on Python 2 vs Python 3 with regards to exceptions, should a method be implemented as an descriptor. In practice this probably makes no difference whatsoever, it's just the same pattern used for forwarding release.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@
 
 - Drop support for Python 3.3.
 
+- The ``HistoricalStorageAdapter`` forwards the ``release`` method to
+  its base instance. See `issue 78 <https://github.com/zopefoundation/ZODB/issues/788>`_.
 
 5.2.4 (2017-05-17)
 ==================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,7 @@
 
 - Drop support for Python 3.3.
 
-- The ``HistoricalStorageAdapter`` forwards the ``release`` method to
+- Ensure that the ``HistoricalStorageAdapter`` forwards the ``release`` method to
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -9,6 +9,8 @@
 
 - Drop support for Python 3.3.
 
+- The ``HistoricalStorageAdapter`` forwards the ``release`` method to
```
tseaver: Prefer imperative voice, e.g.:

```rst
- Ensure that the ``HistoricalStorageAdapter`` forwards the ``release`` method to
```

jamadden: Got it! 

This changelog isn't particularly consistent.

src/ZODB/mvccadapter.py
```diff
@@ -27,10 +27,9 @@ def __init__(self, storage):
 
     def __getattr__(self, name):
         if name in self._copy_methods:
-            if hasattr(self._storage, name):
-                m = getattr(self._storage, name)
-                setattr(self, name, m)
-                return m
+            m = getattr(self._storage, name)
```
jamadden: For the record, I note that I think this is nearly essentially equivalent. Both previous `if` statements, if false, will fall through to the end of the method, which raises `AttributeError`. So the `getattr` here will either raise `AttributeError` (with generally a more helpful error message), or it may raise some other exception that would previously have been masked by `hasattr` in Python 2 (extremely rarely).

jamadden:
Thanks!

Issue #18, opened 2014-03-01
ZODB.serialize.ObjectWriter should fall back to __reduce__() right?


fried:
I know I am trying some 'black magic', but I am creating dynamic classes. I would like to store these in ZODB. These run-time classes implement `__getnewargs__` and `__reduce__` as well as subclass Persistent.  Pickle will dump and load them, because of `__reduce__` providing a viable callable to reproduce these dynamic classes. 

ZODB fails with PicklingError because the type(obj) is not the same as the global
Now I have gotten it to work using a dirty workaround

``` python
from ZODB.serialize import ObjectWriter
orig_serialize = ObjectWriter.serialize
def _serialize(self, obj):
    try:
        return orig_serialize(self, obj)
    except:
        reduce = obj.__reduce__()  # contains (callable/class, newargs, __getstate__() )
        if reduce[1]:
            meta = reduce[0], reduce[1]
        else:
            meta = reduce[0]
        return self._dump(meta, reduce[2])
ObjectWriter.serialize = _serialize
```

Now my dirty workaround works, but I would much rather have the existing serialize method have this functionality

jamadden:
Pickle protocol 1, currently in use in Python 2, doesn't call `__getnewargs__`, for that we'd need protocol 2+ (see #179 ):
```pycon
>>> class X(object):
...     def __getnewargs__(self):
...       raise Exception("getnewargs")
...
>>> import cPickle
>>> cPickle.dumps(X(), 1)
'ccopy_reg\n_reconstructor\nq\x01(c__main__\nX\nq\x02c__builtin__\nobject\nq\x03NtRq\x04.'
>>> cPickle.dumps(X(), 2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in __getnewargs__
Exception: getnewargs
```

jimfulton:
@fried 

Are these the same as persistent classes?

Perhaps take a look at src/ZODB/persistentclass.*

What version of Python are you using?

jamadden:
There've been no comments from the OP since 2014, and the pickle protocol has been updated which theoretically solves this, so I'm going to go ahead and close it.

PR #179, opened 2017-08-26
Use a higher pickle protocol for serializing objects on Python 2

jamadden:
Previously protocol 1 was used. The higher protocol is more efficient for new-style classes (all persistent objects are new-style), according to the docs, at the cost of being very slightly less space efficient for old-style classes.

In tests of a persistent object with two trivial numeric attributes, the higher protocol was 12 bytes smaller, and serialized and deserialized 1us faster. Introducing a reference to another new-style
class (with a small dict and list of strings for attributes) for a more realistic test made the higher protocol twice as fast to serialize (20.5 vs 10.3us), almost half the size (215 vs 142 bytes), and it deserialized 30% faster (6.5 vs 4.6us).

On Python 2, this will now allow open ``file`` objects to be pickled (loading the object will result in a closed file); previously this would result in a ``TypeError`` (as does under Python 3). We had tests that you couldn't do that with a BlobFile so I had to update it to still make that true.

I wouldn't recommend serializing arbitrary open files under Python 2 (for one thing, they can't trivially be deserialized in Python 3), but I didn't take any steps to prevent it either. Since this hasn't
been possible, there shouldn't be code in the wild that is trying to do it---and it wouldn't be forward compatible with Python 3 either.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,17 @@
 - Ensure that the ``HistoricalStorageAdapter`` forwards the ``release`` method to
   its base instance. See `issue 78 <https://github.com/zopefoundation/ZODB/issues/788>`_.
 
+- Use a higher pickle protocol (3) for serializing objects on Python
+  2; previously protocol 1 was used. This is *much* more efficient for
+  new-style classes (all persistent objects are new-style), at the
+  cost of being very slightly less efficient for old-style classes.
+
+  .. note:: On Python 2, this will now allow open ``file`` objects
+			(but **not** open blobs or sockets) to be pickled (loading
+			the object will result in a closed file); previously this
+			would result in a ``TypeError``. Doing so is not
+			recommended as they cannot be loaded in Python 3.
+
 5.2.4 (2017-05-17)
 ==================
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,8 @@
 			would result in a ``TypeError``. Doing so is not
 			recommended as they cannot be loaded in Python 3.
 
+  See `issue 179 <https://github.com/zopefoundation/ZODB/pull/179>`_.
+
 5.2.4 (2017-05-17)
 ==================
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,7 +12,7 @@
 - Ensure that the ``HistoricalStorageAdapter`` forwards the ``release`` method to
   its base instance. See `issue 78 <https://github.com/zopefoundation/ZODB/issues/788>`_.
 
-- Use a higher pickle protocol (3) for serializing objects on Python
+- Use a higher pickle protocol (2) for serializing objects on Python
   2; previously protocol 1 was used. This is *much* more efficient for
   new-style classes (all persistent objects are new-style), at the
   cost of being very slightly less efficient for old-style classes.
--- a/src/ZODB/_compat.py
+++ b/src/ZODB/_compat.py
@@ -34,7 +34,7 @@ if not PY3:
     HIGHEST_PROTOCOL = cPickle.HIGHEST_PROTOCOL
     IMPORT_MAPPING = {}
     NAME_MAPPING = {}
-    _protocol = 1
+    _protocol = 3
     FILESTORAGE_MAGIC = b"FS21"
 else:
     # Python 3.x: can't use stdlib's pickle because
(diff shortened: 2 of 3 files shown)


jamadden:
Another advantage would be that `__getnewargs__` would be called on both Python 2 and 3. This would (probably) solve #18.

jimfulton:
Python 2 pickle/cPickle don't support protocol 3, so this change more closely ties us to zodbpickle, which is a maintenance burden that I'd like to get rid of at some point.

jamadden:
Protocol 2 would have the same benefit, so that could be changed if you'd like.

I just thought it was nice to be able to use the same protocol under both versions, especially since eventually Python 2 will go away. (FWIW, it seems like under Python 2, other than the version number, protocol 2 and 3 are the same...that was why zodbpickle's `binary` object was needed in the first place, wasn't it, to be able to make the distinction under Python 2 that Python 3's protocol 3 makes natively?)

Regarding getting rid of zodbpickle entirely, I would still eventually like to get around to porting protocol 4 back to it, which would make large objects *much* more efficient. And it would be useful for Python 2 and 3.3...which I now remember we just dropped, so that's a bit less important.

jamadden:
D'oh! I forgot the main reason we're committed to `zodbpickle`: Python's `noload` is broken, which breaks parts of ZODB (e.g., `zodbdgc`).

I suppose we could try to get my patch that fixed `noload` for `zodbpickle` upstreamed (I don't *think* it reintroduces the regression that caused its removal), but the earliest that would happen would be 3.7.

jimfulton:
On Sat, Aug 26, 2017 at 10:36 AM, Jason Madden <notifications@github.com>
wrote:

> D'oh! I forgot the main reason we're committed to zodbpickle: Python's
> noload is broken, which breaks parts of ZODB (e.g., zodbdgc).
>
Yes. That was my fault. I regret it. :(

> I suppose we could try to get my patch that fixed noload for zodbpickle
> upstreamed (I don't *think* it reintroduces the regression that caused
> its removal), but the earliest that would happen would be 3.7.
>
Noload was a hack.  The purpose of no-load is to get persistent
references.  I don't see noload being accepted upstream. I *do* see a
possibility for an optimized API to get references. That would probably be
a useful thing to to.

In might be interesting to have a stab at an API to extract references
based on pickletools.genops.  It would likely allow an implementation that
avoids a number of the allocations/instantiations performed by a regular
pickle load.

I should note, however, that noload is an optimization.  If using something
like relstorage or zc.zodbdgc, that optimization is less important because
the computation is performed outside of the database server and doesn't
block anything else, so I consider this optimization less important than I
once did.

jamadden:
> I should note, however, that noload is an optimization.  If using something
like relstorage or zc.zodbdgc, that optimization is less important because
the computation is performed outside of the database server and doesn't
block anything else, so I consider this optimization less important than I
once did.

True, it's an optimization. It's a substantial optimization, the last time I measured. IIRC it was something like an order of a magnitude (but that was 2014 so I may be misremembering). It also is extremely helpful if not everything can be unpickled, either because the class has gone missing or because it relies on infrastructure that's not set up without loading a full application (e.g., zc.baseregistry which unpickles by doing `component.getUtility(IRegistry, name='somename')`.

Another way to accomplish that optimization would be cool!

So, given all that about our ties to zodbpickle, but also the fact that on Python 2, the protocols are pretty much the same, I am happy to either leave it or change it, it's up to you.

jimfulton:
I would prefer to use protocol 2 on Python 2.  This means that the records would remain readable with standard pickle on Python 2, which I think has value.

jamadden:
Makes sense. Done.

jamadden:
Is there anything else I can do to get this ready to be approved?

jmuchemb:
Update the changelog ? (it still describes a change for protocol 3)

jamadden:
> Update the changelog ? (it still describes a change for protocol 3)

Thanks! I have that locally, I just haven't pushed it yet.

jamadden:
Thank you!

jimfulton:
Thank you! :)

jamadden:
As it happens, we're just now hitting a case where protocol 2 might be useful (for `__getnewargs__`). Could there be a PyPI release when convenient please?

jimfulton:
Released.

jamadden:
> On Aug 30, 2017, at 07:55, Jim Fulton <notifications@github.com> wrote:
> 
> Released

Thank you!

jimfulton:
I'm an idiot.  Using protocol 2 defeats zodbpickle's binary mechanism. :(  Whimper.

icemac:
See #193 as followup of this ticket.

PR #181, opened 2017-09-21
Review Connection/DB closure

jmuchemb:
Some parts of the code to close Connection/DB and release resources looks too complicated for me. At least, there is questionable behaviour and also some possible micro optimization. Above all, I've started to look at this for some other change that would depend on it.

### DIFF
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -283,8 +283,7 @@ class Connection(ExportImport, object):
             raise ConnectionStateError("Cannot close a connection joined to "
                                        "a transaction")
 
-        if self._cache is not None:
-            self._cache.incrgc() # This is a good time to do some GC
+        self._cache.incrgc() # This is a good time to do some GC
 
         # Call the close callbacks.
         if self.__onCloseCallbacks is not None:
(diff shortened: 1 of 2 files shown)


src/ZODB/Connection.py
```diff
@@ -951,7 +945,7 @@ def _release_resources(self):
                 c._storage.release()
             c._storage = c._normal_storage = None
             c._cache = PickleCache(self, 0, 0)
-            c.transaction_manager = None
+            c.close(False)
```
jmuchemb: This is 1 of the 3 main changes. I didn't plan to change this at the beginning but while running tests, I discovered that closing a DB left Connections registered to the global transaction manager. Before, these Connections were ignored by patching their afterCompletion/newTransaction.

src/ZODB/DB.py
```diff
@@ -646,15 +646,15 @@ def close(self):
...
-        def _(c):
-            if c.transaction_manager is not None:
-                c.transaction_manager.abort()
-            c.afterCompletion = c.newTransaction = c.close = noop
-            c._release_resources()
+        def _(conn):
+            if conn.transaction_manager is not None:
+                for c in six.itervalues(conn.connections):
+                    c.explicit_transactions = True
+                conn.transaction_manager.abort()
```
jmuchemb: The 2 other big changes:
- Make sure that the real `newTransaction` is not called when aborting an existing transaction. This was not the case if `explicit_transactions=0`.
- As written above, Connections of closed DB could remain registered, and when it happened, multi-db was probably broken because only the primary connections were ignored.

Setting `explicit_transactions=1` is also simpler than patching newTransaction.

jmuchemb:
I have rebased and I wrote a commit message.

Unless there's any objection, I'll merge on Friday.

> Above all, I've started to look at this for some other change that would depend on it.

Then I'll open a merge request to process invalidations immediately for closed connections, mainly to avoid memory leaks when pool-timeout is not set.

jimfulton:
Yay simpler!!! Sorry I haven't looked at this. I'll look at this this weekend.

jimfulton:
This looks great. Thanks!  I added a comment for `c.explicit_transactions = True`.   When CI is green, I'll merge.

jimfulton:
Oh, WIP? Is this still a work in progress? Or is it ready for merge?

PR #182, opened 2017-10-27
Unpin specific PyPy versions, blacklist 3.3 in `setup.py`

Rotonen:
Unpin specific PyPy versions, blacklist 3.3 in `setup.py` as Travis now offers a new enough PyPy 3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ sudo: false
 matrix:
     include:
         - os: linux
-          python: pypy-5.6.0
+          python: pypy
         - os: linux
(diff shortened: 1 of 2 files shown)


jamadden:
The default version of PyPy is now 5.8.0/2.7.13 and the default PyPy3 is 5.8.0-beta/3.5.3. I imagine those should be good enough for the foreseeable future (they won't go backwards, only forwards). Most other zopefoundation projects have already stopped or are stopping the use of particular version pins for PyPy now that the update has taken place. It would be good to avoid introducing the complexity of a manual pyenv if we don't need it.

PR #184, opened 2017-11-07
Report full oid of missing object during pack

davisagli:
@hvelarde had an issue with a db that couldn't be packed due to an object that could not be found. This makes sure that the entire oid is reported in that case, instead of just the final 2 bytes.

Or would it be better to handle this in `fsIndex`'s `__getitem__`?


hvelarde:
thanks, @davisagli!

PR #186, opened 2017-11-14
Update the front page to be a little less negative.

jimfulton:
Especially with regard to write capacity and search.

This hasn't been updated in a while and I feel that it is out of date.

I feel that this could use more work, but I'm anxious to do an
incremental improvement given the screencast I just released. :)


freddrake:
+1

PR #188, opened 2018-02-10
Avoid seeking on import

jimfulton:
I'm working on a project where I need to copy object trees that contain many (10s of thousands) of large blobs.  The easiest way to do this (for my application so far) is to leverage export/import.  I wanted to avoid extra file I/O, so I tried exporting to and importing from a pipe.  This failed because import looks for blob markers and then seeks back when they're not found, which is most of the time.  The blob marker (and associated seek) wasn't necessary because the blob object record can serve as a marker.

The primary change in this PR is to leverage blob object records as blob data markers, thus avoiding seeks and permitting pipes to be used.  For backward compatibility, blob-data markers are still written on output and read and checked on input.

In addition to allowing pipes to be used, avoiding seeks should improve performance, because seeks cause file buffers to be discarded.

This PR also provides 2 ancillary blob-related performance improvements:

- When writing blob data during import, temporary files are written to the storage's blob temporary directory, assuring that we don't have to make extra copies.

- The default buffer size for ZODB.utils.cp was increased to (still modest) 64K and made overridable.  Export was updated to allow a larger buffer size to be passed in. (There was no obvious way to make the import buffer size overridable, due to the way things get called.)

### DIFF
--- a/src/ZODB/ExportImport.py
+++ b/src/ZODB/ExportImport.py
@@ -158,9 +158,15 @@ class ExportImport(object):
                 oids[ooid] = oid = self._storage.new_oid()
                 return_oid_list.append(oid)
 
-            # Blob support
-            blob_begin = f.read(len(blob_begin_marker))
-            if blob_begin == blob_begin_marker:
+            if (len(data) < 99 and 'blob' in data and
+                isinstance(self._reader.getGhost(data), Blob)
+                ):
+                # Blob support
+
+                # Make sure we have a (redundant, overly) blob marker.
+                if f.read(len(blob_begin_marker)) != blob_begin_marker:
+                    raise ValueError("No data for blob object")
+
                 # Copy the blob data to a temporary file
                 # and remember the name
                 blob_len = u64(f.read(8))
@@ -169,7 +175,6 @@ class ExportImport(object):
                 cp(f, blob_file, blob_len)
                 blob_file.close()
             else:
-                f.seek(-len(blob_begin_marker),1)
                 blob_filename = None
 
             pfile = BytesIO(data)
--- a/src/ZODB/ExportImport.py
+++ b/src/ZODB/ExportImport.py
@@ -31,7 +31,7 @@ logger = logging.getLogger('ZODB.ExportImport')
 
 class ExportImport(object):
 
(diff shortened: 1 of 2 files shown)


src/ZODB/ExportImport.py
```diff
@@ -31,7 +31,7 @@
 
 class ExportImport(object):
 
-    def exportFile(self, oid, f=None):
+    def exportFile(self, oid, f=None, bufsize=1 << 16):
```
mgedmin: I would find `64*1024` more readable than `1 << 16`.

src/ZODB/ExportImport.py
```diff
@@ -158,18 +158,23 @@ def persistent_load(ooid):
                 oids[ooid] = oid = self._storage.new_oid()
                 return_oid_list.append(oid)
 
-            # Blob support
-            blob_begin = f.read(len(blob_begin_marker))
-            if blob_begin == blob_begin_marker:
+            if (len(data) < 99 and b'blob' in data and
```
mgedmin: 99 seems a bit magic.  I'd like to see it defined as a named constant at the top level, with a comment explaining why a pickled blob will never be longer than that.

src/ZODB/utils.py
```diff
@@ -95,7 +95,7 @@ def u64(v):
 U64 = u64
 
 
-def cp(f1, f2, length=None):
+def cp(f1, f2, length=None, bufsize=1 << 16):
```
mgedmin: `64 * 1024`

src/ZODB/ExportImport.py
jimfulton: 99 is arbitrary (it's a Jim's number ;-) ).  It's chosen to be bigger than I'd expect a blob record to ever be. It's a micro optimization to avoid searching for the test 'blob' in a larger pickle.  Maybe we should just drop it.

I hate having to look at the top of a file to see what something means.  (That's why I dislike from imports.)

I'll just remove it.

src/ZODB/ExportImport.py
jimfulton: Funny, I find 1 << 16 more readable.  It's practically a literal to me.

PR #194, opened 2018-03-25
Save object ids as zodbpickle.binary and use pickle protocol 3 for Python 2

### LINKED ISSUE #193: We need to use pickle protocol 3 for Python 2 and store oids as zodbpickle.binary
Under Python 2:

We should be storing oids as zodbpickle.binary so they are treated as binary when read by Python 3.

We use pickle protocol 2 under Python 3, however, which defeats zodbpickle.binary, so we need to use protocol 3.

jimfulton:
Fixes #193

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,19 @@
 5.4.0 (unreleased)
 ==================
 
+- ZODB now uses pickle protocol 3 for both Python 2 and Python 3.
+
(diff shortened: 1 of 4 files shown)


jimfulton:
Thanks.  WRT OIDs, I went through the same thought process. :)

jamadden:
I can confirm that RelStorage tests pass with this change.

jimfulton:
Thanks!

PR #199, opened 2018-03-28
Fix infinite recursion when copying from a MappingStorage to a FileStorage

MatthewWilkes:
This never would have worked, there's an obvious typo, and it bit me today when trying to persist a stack of DemoStorages in a test run, for debugging.


jimfulton:
Thanks!

Issue #203, opened 2018-05-14
Possible memory leak using python 2.7, zodb 5 and cacheMinimize


azmeuk:
After upgrading from zodb 4.4.5 to zodb 5.4.0 I experienced an increasing memory consumption in a daemon process I developed. At some point it saturated the whole memory in my server so I tried to understand what was going on, and I managed to reproduce the memory growth with this script:

```python
import transaction
import ZODB
from BTrees.OOBTree import OOBTree

conn = ZODB.DB(None).open()
conn.root()["foobar"] = OOBTree()
transaction.commit()

for i in range(10 ** 7):
    conn.root()["foobar"]
    conn.cacheMinimize()

conn.close()
```

If you run this script with python 2.7 and zodb 5 the memory consumption of the python process will grow slowly but steadily. The size of the chunks I observed was sometimes 256kB, sometimes 4kB, but I do not really know how to reproduce specifically a growth by 256 or 4kB.

If you comment the line `conn.cacheMinimize()` or the line `conn.root()["foobar"]` the memory usage will not grow.

I [wrote a proof of concept](https://github.com/azmeuk/zodbleak) that generates a memory consumption chart comparison :
![Memory usage](https://raw.githubusercontent.com/azmeuk/zodbleak/master/results/eloi-archlinux.png)

I also asked other developers at my company to test how the memory consumption behave on their computers with python 2.7, 3.4, 3.5, 3.6 and zodb 4.4.5 and zodb 5.4.0. The only case that always grows is python 2 and zodb 5. You can see the results [here](https://github.com/azmeuk/zodbleak/tree/master/results), with the Linux distributions used. We did not test on Mac or Windows.

jamadden:
I can reproduce the memory growth exactly as described with ZODB 5.4.0 and CPython 2.7.15 on macOS.

Adding a `gc.collect()` call to the loop stops the memory growth, so it is most likely a reference cycle. Using a call to `objgraph.show_growth()` (that has been patched to *not* call `gc.collect`) shows one new tuple for each iteration of the loop.

jamadden:
I was able to simplify the case I was observing. The pair `conn.root()['foobar']` and `conn.cacheMinimize()` turns out to basically be equivalent to `db_root = conn.root(); conn.setstate(db_root)`. If we inline as many intermediate uses of `conn` and `conn._reader` as we can, we arrive at something like this:

```python
from ZODB._compat import PersistentUnpickler
def load_persistent(arg):
    return None

[...]

PR #205, opened 2018-05-16
Bump dependency on zodbpickle to at least 1.0.1

jamadden:
This is required to avoid a memory leak on Python 2.7.

See #203.


jimfulton:
Thanks!

PR #207, opened 2018-05-31
Make TransactionMetaData in charge of (de)serializing extension data

jmuchemb:
IStorage implementations used to do this task themselves which leads to code
duplication and sometimes bugs (one was fixed recently in NEO). Like for object
serialization, this should be done by the upper layer (Connection).

This commit also provides a way to get raw extensions data while iterating
over transactions (this is actually the original purpose[2]). So far, extension
data could only be retrieved unpickled, which caused several problems:

- tools like `zodb dump` [1] cannot dump data exactly as stored on a
  storage. This makes database potentially not bit-to-bit identical to
  its original after restoring from such dump.

- `zodb dump` output could be changing from run to run on the same
  database. This comes from the fact that e.g. python dictionaries are
  unordered and so when pickling a dict back to bytes the result could
  be not the same as original.

  ( this problem can be worked-around partly to work reliably for e.g.
    dict with str keys - by always emitting items in key sorted order,
    but it is hard to make it work reliably for arbitrary types )

Both issues make it hard to verify integrity of database at the lowest
possible level after restoration, and make it hard to verify bit-to-bit
compatibility with non-python ZODB implementations.

For this, TransactionMetaData gets a new 'extension_bytes' attribute and
and common usage becomes:

* Application committing a transaction:

  - 'extension' is set with a dictionary
  - the storage gets the bytes via 'extension_bytes'

* Iteration:

  - the storage passes bytes as 'extension' parameter of TransactionMetaData
  - the application can get extension data either as bytes ('extension_bytes')
    or deserialized ('extension'): in the former case, no deserialization
    happens and the returned value is exactly what was passed by the storage

[1] https://lab.nexedi.com/nexedi/zodbtools
[2] https://github.com/zopefoundation/ZODB/pull/183

Co-Authored-By: Kirill Smelkov <kirr@nexedi.com>

### DIFF
--- a/src/ZODB/Connection.py
+++ b/src/ZODB/Connection.py
@@ -49,10 +49,10 @@ from ZODB.utils import p64, u64, z64, oid_repr, positive_id
 from ZODB import utils
 import six
 
+from ._compat import dumps, loads, _protocol
 from .mvccadapter import HistoricalStorageAdapter
 
 from . import valuedoc
-from . import _compat
 
 global_reset_counter = 0
 
@@ -1303,10 +1303,24 @@ large-record-size option in a configuration file) to specify a larger
 size.
 """
 
+class overridable_property(object):
+    """
+    Same as property() with only a getter, except that setting a
+    value overrides the property rather than raising AttributeError.
+    """
+
+    def __init__(self, func):
+        self.__doc__ = func.__doc__
+        self.func = func
+
+    def __get__(self, obj, cls):
+        return self if obj is None else self.func(obj)
+
+
 @implementer(IStorageTransactionMetaData)
 class TransactionMetaData(object):
 
-    def __init__(self, user=u'', description=u'', extension=b''):
+    def __init__(self, user=u'', description=u'', extension=None):
         if not isinstance(user, bytes):
             user = user.encode('utf-8')
         self.user = user
@@ -1315,9 +1329,20 @@ class TransactionMetaData(object):
             description = description.encode('utf-8')
         self.description = description
 
-        if not isinstance(extension, dict):
-            extension = _compat.loads(extension) if extension else {}
-        self.extension = extension
+        if isinstance(extension, bytes):
+            self.extension_bytes = extension
+        else:
+            self.extension = {} if extension is None else extension
+
+    @overridable_property
+    def extension(self):
+        extension_bytes = self.extension_bytes
+        return loads(extension_bytes) if extension_bytes else {}
+
+    @overridable_property
+    def extension_bytes(self):
+        extension = self.extension
+        return dumps(extension, _protocol) if extension else b''
 
     def note(self, text): # for tests
         text = text.strip()
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,12 @@
   rolled back. See `issue 268
   <https://github.com/zopefoundation/ZODB/issues/268>`_.
 
+- Make TransactionMetaData in charge of (de)serializing extension data.
+  A new ``extension_bytes`` attribute converts automatically from
+  ``extension``, or vice-versa. During storage iteration, ``extension_bytes``
+  holds bytes as they are stored (i.e. no deserialization happens).
+  See `issue 207 <https://github.com/zopefoundation/ZODB/pull/207>`_.
+
 5.5.1 (2018-10-25)
 ==================
 
--- a/src/ZODB/BaseStorage.py
+++ b/src/ZODB/BaseStorage.py
@@ -31,7 +31,7 @@ from . import POSException, utils
 from .Connection import TransactionMetaData
 from .utils import z64, oid_repr, byte_ord, byte_chr, load_current
 from .UndoLogCompatible import UndoLogCompatible
-from ._compat import dumps, _protocol, py2_hasattr
+from ._compat import py2_hasattr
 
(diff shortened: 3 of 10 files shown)


jmuchemb:
I recently got more interested by #183 while doing 2 things in NEO: check that a FS DB was correctly imported (using zodbtools), and fight with bugs and code duplication about serialization of extension data.

But like Jim, I prefer to serialize between `extension` and the new `extension_bytes` when necessary. I do it lazily with properties.

Since `extension_bytes` is not optional anymore, many comments in #183 become irrelevant and it looks less controversial to define a new attribute IStorageTransactionMetaData, so I still do it.

jamadden:
I think I like this (though I'm not fond of the magic that the decorator uses to get the attribute names from the argspec)!

jmuchemb:
> though I'm not fond of the magic that the decorator uses to get the attribute names from the argspec

I found important to minimize code duplication here, so that both properties remain always symmetric. I wish I could find something better. Despite code that is a little harder to read, there's still little some duplication at the places `_extension_property` is used.

I don't want to wait too long before merging this one, as I have other changes I'd like to submit.

I plan to merge mid-September, unless anyone wants more time to review.

src/ZODB/Connection.py
```diff
@@ -1300,6 +1300,23 @@ def __repr__(self):
 size.
 """
 
+def _extension_property(func):
+    import inspect
```
tseaver: Imports at top-of-module for non-test code please.  Also, this descriptor factory needs a docstring explaining just what it does.

jmuchemb: > Imports at top-of-module for non-test code please.

This function is only called at load-time so I imported locally to not pollute globals. But if you prefer, I'll change.

> this descriptor factory needs a docstring explaining just what it does.

Ouch, not easy. What about:
```py
def _extension_property(func):
    """
    When called twice, this builds a pair of properties that maintain a value
    in 2 different forms, in such a way that conversion between the 2 is done
    lazily. The 2 decorated functions must be inverse of each other:
      def foo(bar): ...
      def bar(foo): ...
      where foo(bar(x)) == x

    The name of their parameter specifies that of the mirror attribute.

    This helps to make sure that 2 properties remain always symmetric.
    Currently used to automatically convert transaction extension from/to bytes.
    """
```

jmuchemb: The main goal of this PR is to avoid (de)serialize whenever possible, because `dumps(loads(x))` is not always equal to `x` and we want to be able to compare bit-to-bit (longer explanation in the PR description).

tseaver: If `dumps` and `loads` are not complements, then the whole use case blows up.  Store the data on disk in one format, sure, but keep it in only one *real* attribute while in RAM, and let the other one be computed logically from it (I don't really care whether `extension` or `extension_bytes` is the canonical one.

jmuchemb: > If `dumps` and `loads` are not complements, then the whole use case blows up.

I don't understand.

> I don't really care whether `extension` or `extension_bytes` is the canonical one.

If `extension_bytes` is the canonical one, yes it works. But from a performance point of view, that looks horrible, since you'd convert whenever you access/modify `extension` (isn't it the main way extensions are manipulated ?)

jamadden: Couldn't this be broken anyway? `extension` is a `dict`, and I would guess if there's any mutation of it, it's in-place, not by setting a whole new `dict` on the property.

jmuchemb: Arg, I missed that. The only way to solve that while still storing extension_bytes verbatim, is to delete the cached extension_bytes whenever extension is accessed. I must change `_extension_property`.

jmuchemb: I mean:
```diff
@@ -1309,7 +1322,7 @@ def fget(self):
             return self.__dict__[name]
         except KeyError:
             d = self.__dict__
-            value = d[name] = func(d[alt])
+            value = d[name] = func(d.pop(alt))
             return value
     def fset(self, value):
         d = self.__dict__
```

jamadden: Sorry, now that I'm thinking about it, that still seems a bit foot-gunny to me. If the intent is to be able to copy a database using the iterator APIs byte-for-byte identical, then if *anything* in that process touches `.extension` we'd have to reserialize, and that would defeat the purpose. 

Maybe a big-ole disclaimer in the docs is enough to mitigate that?

jmuchemb: But code doing bit-to-bit checks is unlikely to access `.extension`, isn't it ?

With my suggested fixup, the PR preserves the original behaviour, in that any change to .extension is "saved" (i.e. serialized), but that also does not look important: I'd be surprised to see code relying on that.

My feeling is that either behaviour is fine, but whatever we choose (between too much or not enough reserialization), I suppose we can't avoid documenting.

icemac:
What is the status of this PR? What needs to be done to get it merged?

jmuchemb:
First, I rebased the original commit. But even with the small diff I suggested above, it would remain possible to mutate extensions after having kept a reference to them.

Magic does not look a good idea if the result is not perfect, especially when there's no need.

So I changed for something simpler, following [tseaver comment](#discussion_r216415721) partially. See `IStorageTransactionMetaData.__doc__` for more details.

jmuchemb:
I'll merge next monday.

PR #222, opened 2018-08-23
Avoid most calls to time.sleep() in the tests by mocking time.[gm]time

jamadden:
This also makes us more deterministic. (Based on [code I wrote for nti.testing](https://github.com/NextThought/nti.testing/blob/master/src/nti/testing/time.py).)

On my machine, this speeds up a sequential test run from 4:08 to 1:30. An AppVeyor 2.7, where we run tests sequentially, the time went from over five minutes to 2:35.

We no longer get much benefit from running the layer tests in parallel on CI, in fact it looks like the overhead actually slows them down slightly, so stop doing that.

This obsoletes the hack we have for running coverage in parallel on CI; I plan a followup PR to remove that (if this is accepted). 

The tests fail on appveyor 3.4 because its version of setuptools is so old. I would like to also address that in the same PR that adjusts the coverage hacks on Travis. (EDIT: Addressed when I added the lock.)

### DIFF
--- a/src/ZODB/tests/testblob.py
+++ b/src/ZODB/tests/testblob.py
@@ -18,11 +18,7 @@ from ZODB.FileStorage import FileStorage
 from ZODB.tests.testConfig import ConfigTestBase
 from ZODB._compat import Pickler, Unpickler, _protocol
 
-import os
-if os.environ.get('USE_ZOPE_TESTING_DOCTEST'):
-    from zope.testing import doctest
-else:
-    import doctest
+import doctest
 
 import os
 import random
@@ -40,12 +36,7 @@ import ZODB.tests.StorageTestBase
 import ZODB.tests.util
 import zope.testing.renormalizing
 
-
-try:
-    from StringIO import StringIO as BytesIO
-except ImportError:
-    # Py3
-    from io import BytesIO
+from io import BytesIO
 
 try:
     file_type = file
@@ -67,7 +58,11 @@ def new_time():
     now = new_time = time.time()
     while new_time <= now:
         new_time = time.time()
-    time.sleep(1)
+    if time.time() - new_time < 1.0:
+        # Detect if we're in a time monotonically increasing
+        # layer (two back-to-back calls of time.time() advance the clock
+        # by a whole second); if so, we don't need to sleep
+        time.sleep(1.0)
     return new_time
 
 
@@ -703,6 +698,14 @@ def setUp(test):
     ZODB.tests.util.setUp(test)
     test.globs['rmtree'] = zope.testing.setupstack.rmtree
 
+def timeIncreasesSetUp(test):
+    setUp(test)
+    l = test.globs['time_layer'] = ZODB.tests.util.MonotonicallyIncreasingTimeMinimalTestLayer('')
+    l.testSetUp()
+
+def timeIncreasesTearDown(test):
+    test.globs['time_layer'].testTearDown()
+    util.tearDown(test)
 
 def setUpBlobAdaptedFileStorage(test):
     setUp(test)
@@ -791,7 +794,7 @@ def storage_reusable_suite(prefix, factory,
     if test_undo:
         add_test_based_on_test_class(BlobUndoTests)
 
-    suite.layer = ZODB.tests.util.MininalTestLayer(prefix+'BlobTests')
+    suite.layer = ZODB.tests.util.MonotonicallyIncreasingTimeMinimalTestLayer(prefix+'BlobTests')
 
     return suite
 
@@ -804,9 +807,15 @@ def test_suite():
         "blob_basic.txt",
         "blob_consume.txt",
         "blob_tempdir.txt",
-        "blobstorage_packing.txt",
         setUp=setUp,
         tearDown=util.tearDown,
+        optionflags=doctest.ELLIPSIS,
+        checker=ZODB.tests.util.checker,
+        ))
+    suite.addTest(doctest.DocFileSuite(
+        "blobstorage_packing.txt",
+        setUp=timeIncreasesSetUp,
+        tearDown=timeIncreasesTearDown,
         optionflags=doctest.ELLIPSIS,
         checker=ZODB.tests.util.checker,
         ))
--- a/src/ZODB/tests/util.py
+++ b/src/ZODB/tests/util.py
@@ -29,6 +29,19 @@ from ZODB.Connection import TransactionMetaData
 import zope.testing.setupstack
 from zope.testing import renormalizing
 
+try:
+    from unittest import mock
+except ImportError:
+    import mock
+
+import six
+import functools
+from time import time as _real_time
+from time import gmtime as _real_gmtime
+_current_time = _real_time()
+
+
+
 checker = renormalizing.RENormalizing([
     (re.compile("<(.*?) object at 0x[0-9a-f]*?>"),
      r"<\1 object at 0x000000000000>"),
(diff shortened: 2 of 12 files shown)


jamadden:
I see one legit failure across all the runs:
```

374Error in test checkPackLotsWhileWriting (ZODB.tests.testDemoStorage.DemoStorageHexTests)
375Traceback (most recent call last):
376  File "c:\python36-x64\lib\unittest\case.py", line 59, in testPartExecutor
377    yield
378  File "c:\python36-x64\lib\unittest\case.py", line 605, in run
379    testMethod()
380  File "c:\projects\zodb\src\ZODB\tests\util.py", line 253, in wrapper
381    return func(*args, **kwargs)
382  File "c:\projects\zodb\src\ZODB\tests\PackableStorage.py", line 316, in checkPackLotsWhileWriting
383    db.pack(packt)
384  File "c:\projects\zodb\src\ZODB\DB.py", line 854, in pack
385    self.storage.pack(t, self.references)
386  File "c:\projects\zodb\src\ZODB\tests\hexstorage.py", line 65, in pack
387    return self.base.pack(pack_time, refs, gc)
388  File "c:\projects\zodb\src\ZODB\DemoStorage.py", line 318, in pack
389    return self.changes.pack(t, referencesf, gc=gc)
390  File "c:\projects\zodb\src\ZODB\utils.py", line 288, in __call__
391    return func(*args, **kw)
392  File "c:\projects\zodb\src\ZODB\MappingStorage.py", line 199, in pack
393    raise ValueError("Already packed to a later time")
394ValueError: Already packed to a later time
395
396
397 checkPackNowWhileWriting (ZODB.tests.testDemoStorage.DemoStorageHexTests) (0.109 s)
398
```
That's a test that uses threads, so I suspect I need a lock.

src/ZODB/tests/testblob.py
```diff
@@ -60,14 +51,17 @@ def new_time():
...
 
     """
     now = new_time = time.time()
     while new_time <= now:
         new_time = time.time()
-    time.sleep(1)
+    if time.time() - new_time >= 0.5:
+        # Detect if we're in a time monotonically increasing
+        # layer; if so, we don't need to sleep
+        time.sleep(0.5)
```
jmuchemb: I don't understand. For me, `time.time() - new_time >= 0.5` is usually false when using real time.

Anyway, I think we always want to sleep. At least more than the resolution of the FS/kernels we want to support. Or is it for another reason (I admit I don't know blobs and blob tests enough) ?

src/ZODB/tests/util.py
```diff
@@ -206,3 +219,117 @@ def clear_transaction_syncs():
...
+            global _current_time # pylint:disable=global-statement
+            with self._lock:
+                _current_time = max(_real_time(), _current_time + self._granularity)
+            return _current_time
+        fake_time.side_effect = incr
+
+        def incr_gmtime(*seconds):
+            if seconds:
+                assert len(seconds) == 1
+                now = seconds[0]
```
jmuchemb: What about replacing the above 2 lines with `now, = seconds`.

src/ZODB/tests/util.py
```diff
@@ -206,3 +219,117 @@ def clear_transaction_syncs():
...
+    _current_time = value
+
+
+class MonotonicallyIncreasingTimeMinimalTestLayer(MininalTestLayer):
+
+    def testSetUp(self):
+        self.fake_gmtime = mock.Mock()
+        self.fake_time = mock.Mock()
+        self.time_manager = _TimeWrapper(0.1)
+        self.time_manager.configure_fakes(self.fake_gmtime, self.fake_time)
```
jmuchemb: Why storing fake_gmtime, fake_time and time_manager on self ?

src/ZODB/tests/testblob.py
jamadden: Sleeping was done just to advance the `time.time()` counter, because that's what we create TIDs and pack based on. I don't *think* filesystem times are relevant at all. If the `time.time()` counter is guaranteed to have advanced by the pack resolution, we don't need to sleep.

src/ZODB/tests/testblob.py
```diff
@@ -60,14 +51,17 @@ def new_time():
     """Create a _new_ time stamp.
 
     This method also makes sure that after retrieving a timestamp that was
-    *before* a transaction was committed, that at least one second passes so
+    *before* a transaction was committed, that at least one half second passes so
```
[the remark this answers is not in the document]
jamadden: I can revert the comment, the switch to a finer resolution is a leftover from an earlier iteration.

src/ZODB/tests/util.py
jamadden: I don't like relying on the side-effect of iterable unpacking that way.

src/ZODB/tests/util.py
jamadden: Ah, that's a half-finished thought. I will adjust that as well.

src/ZODB/tests/testblob.py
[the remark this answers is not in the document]
jamadden: My guess is that the comment about platfroms means the same as the comment in [`TransactionalUndoStorage.snooze`](https://github.com/zopefoundation/ZODB/blob/master/src/ZODB/tests/TransactionalUndoStorage.py#L41): "In Windows, it's possible that two successive time.time() calls return the same value."

src/ZODB/tests/testblob.py
jamadden: > You didn't answer about the fact that the condition (>= 0.5) is inconsistent with the comment.

True, I just fixed it 😄

src/ZODB/tests/util.py
jmuchemb: or `(now,) = seconds`

jamadden: How about a kwarg? `def incr_gmtime(seconds=None): if seconds:` It can still be passed positionally, and because the C version doesn't allow kwargs at all (it throws a TypeError; there's no way to spell optional-but-only-positional in pure Python AFAIK) it remains compatible.

jmuchemb: A default parm value is also fine for me (I dislike the `assert len(foo) == 1; bar = foo[0]` patten so much).

jamadden:
And the issue half resolved itself. There is apparently an unrelated one-off error under 3.7.

jamadden:
I can reproduce the failing test locally both on this branch and on master:
```console
$ zope-testrunner --test-path=src -m testCache -t testBogusObject -N 10
Running zope.testrunner.layer.UnitTests tests:
  Set up zope.testrunner.layer.UnitTests in 0.000 seconds.
Iteration 1


Failure in test testBogusObject (ZODB.tests.testCache.CacheErrors)
  File "//3.7/lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "//3.7/lib/python3.7/unittest/case.py", line 615, in run
    testMethod()
  File "//ZODB/src/ZODB/tests/testCache.py", line 381, in testBogusObject
    self.assertEqual(rc(None), nones)
  File "//Versions/3.7/lib/python3.7/unittest/case.py", line 839, in assertEqual
    assertion_func(first, second, msg=msg)
  File "//Versions/3.7/lib/python3.7/unittest/case.py", line 832, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 16371 != 16375
```

I will look at fixing it.

jamadden:
I fixed the 3.7 issue. The reason we didn't see it on Travis is that we run Travis builds with coverage enabled, and that test is disabled when coverage is enabled. The reason we didn't see it on appveyor...it's slightly flaky? 

The 3.7 patch could be cherry-picked to a different PR (and would need to be if this PR isn't accepted). I would like to move on to a different issue, but I would also like the tests to take less time to run when I do so 😄

jamadden:
Thank you!

jimfulton:
This has broken the ZEO tests, presumably because this can't control time on the server. Somewhat surprisingly, only checkPackUndoLog is failing.

Amongst the options:
- Remove the decorator for that test.
- Try to make the decorator a no-op for ZEO via monkey patching.
- Update the decorator to look at some test meta-data to opt out.

@jamadden thoughts?

jamadden:
Oy, thank you for looking into the ZEO failures. I apologize for causing this problem, and I thank you for offering me the opportunity to help fix it. I'll take a look as soon as I can.

jimfulton:
Ha, thanks for all the work you do. No apologies necessary!

jimfulton:
BTW, I wasn't asking you to fix it, just asking for your opinion on the options.

jamadden:
I looked at this a little bit. The ZODB tests are de-facto a public API (something I always forget), extended by ZEO, RelStorage, and probably NEO tests. So it would seem that if we're going to introduce potentially breaking changes here, it would behoove us to provide clients a way to officially adapt to those changes. That argues in favor of having a way for the decorator to be controlled by the client classes (option 3).

However, given that:

- only this one ZEO test case seems affected by this (I double checked the RelStorage tests, they're fine);
- and that it would take a release of ZODB to introduce some new test API;
- and that I personally dislike making test-only releases (where possible), 

I think I'm fine with including a patch for this test directly in the ZEO sources. At least until this becomes a more widespread issue.

The following in `testZEO.FullGenericTests` worked for me (on Python 3; it will take some tweaking to work on Python 2):
```python
    def checkPackUndoLog(self):
        # PackableStorage.PackableUndoStorage wants to adjust time.sleep and time.time
        # to cooperate and pretend for time to pass. That doesn't quite
        # work out for the spawned server.
        super_meth = super(FullGenericTests, self).checkPackUndoLog
        # Find the underlying function, not the decorated method.
        # If it doesn't exist, the implementation has changed and we
        # need to revisit this.
        super_func = super_meth.__wrapped__
        super_func(self)
```

I can open a PR in the ZEO repo with this (assuming I can get it to work with Python 2) if that sounds reasonable.

PR #235, opened 2018-11-26
Repozo : add an option to verify on recovery

Sebatyne:
This PR aims to reduce the number of IO operations when recovering a ZODB backup with repozo, when user wants to also verify the data. The idea is to add a new option for the -R/--recovery mode which will run the same checks as the -v/--Verify mode, but by reading data in increment files only once.

Also, I did a first commit to prevent the concat() function to close the file descriptor, as currently it is closed twice : once in concat(), and once in the function calling concat(). I think it should be the responsability of the one opening the file to also close it. What surprised me is that there was an assertion in tests checking that concat() was indeed closing the file descriptor. I couldn't find in the git history the explanation why.

### DIFF
--- a/src/ZODB/scripts/repozo.py
+++ b/src/ZODB/scripts/repozo.py
@@ -73,6 +73,12 @@ Options for -R/--recover:
         Note:  for the stdout case, the index file will **not** be restored
         automatically.
 
+    -w
+    --with-verification
+        Verify on the fly the backup files on recovering. This option runs
+        the same checks as when repozo is run in -V/--verify mode, and
+        allows to verify and recover a backup in one single step.
+
 Options for -V/--verify:
     -Q / --quick
         Verify file sizes only (skip md5 checksums).
@@ -108,6 +114,8 @@ class WouldOverwriteFiles(Exception):
 class NoFiles(Exception):
     pass
 
+class VerificationFail(Exception):
+    pass
 
 class _GzipCloser(object):
 
@@ -146,7 +154,7 @@ def error(msg, *args):
 def parseargs(argv):
     global VERBOSE
     try:
-        opts, args = getopt.getopt(argv, 'BRVvhr:f:FQzkD:o:',
+        opts, args = getopt.getopt(argv, 'BRVvhr:f:FQzkD:o:w',
                                    ['backup',
                                     'recover',
                                     'verify',
@@ -160,6 +168,7 @@ def parseargs(argv):
                                     'kill-old-on-full',
                                     'date=',
                                     'output=',
+                                    'with-verification',
                                    ])
     except getopt.error as msg:
         usage(1, msg)
@@ -174,6 +183,7 @@ def parseargs(argv):
         quick = False       # -Q flag state
         gzip = False        # -z flag state
         killold = False     # -k flag state
+        withverify = False  # -w flag state
 
     options = Options()
 
@@ -210,6 +220,8 @@ def parseargs(argv):
             options.gzip = True
         elif opt in ('-k', '--kill-old-on-full'):
             options.killold = True
+        elif opt in ('-w', '--with-verify'):
+            options.withverify = True
         else:
             assert False, (opt, arg)
 
@@ -229,6 +241,9 @@ def parseargs(argv):
         if options.output is not None:
             log('--output option is ignored in backup mode')
             options.output = None
+        if options.withverify is not None:
+            log('--with-verify option is ignored in backup mode')
+            options.withverify = None
     elif options.mode == RECOVER:
         if options.file is not None:
             log('--file option is ignored in recover mode')
@@ -256,6 +271,9 @@ def parseargs(argv):
         if options.killold:
             log('--kill-old-on-full option is ignored in verify mode')
             options.killold = False
+        if options.withverify is not None:
+            log('--with-verify option is ignored in verify mode')
+            options.withverify = None
     return options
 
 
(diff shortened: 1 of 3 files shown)


src/ZODB/scripts/repozo.py
```diff
@@ -73,6 +73,12 @@
         Note:  for the stdout case, the index file will **not** be restored
         automatically.
 
+    -w
+    --with-verification
+        Verify on the fly the backup files on recovering. This option runs
+        the same checks as when repozo is run in -V/--verify mode, and
+        allows to verify and recover a backup in one single step.
```
mgedmin: I wonder if it's worth mentioning that if verification fails, the output file is left in a partial, inconsistent state.

Sebatyne:
Hello,

Thank you @icemac and @mgedmin for your good words and your valuable inputs.

I've added an entry in the changelog, and squashed the change into the commit introducing the feature.

As I was at it, I've added a prefix argument to all mkdtemp calls in test_repozo.py, as there weren't numerous.

Regards,

Nicolas

src/ZODB/scripts/repozo.py
Sebatyne: That's an interesting issue you are mentioning. From a user experience point of view, do you think that deleting the partially restored zodb would be a good idea ? Otherwise user may use it (ie: after a disaster recovery) without noticinig data is missing. More especially, this partially restored file is valid according to the zodb format.

src/ZODB/scripts/repozo.py
```diff
@@ -730,6 +776,8 @@ def main(argv=None):
             do_backup(options)
         except WouldOverwriteFiles as e:
             sys.exit(str(e))
+        except VerificationFail as e:
+            sys.exit(str(e))
```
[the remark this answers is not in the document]
Sebatyne: Here do you propose to use base class for all 3 repozo custom errors (WouldOverwriteFiles, NoFiles and VerificationFail), or do you propose to merge some of them ? From my understanding, all these exceptions are not raised because of bugs, but just to inform user that repozo didn't finish correctly, most likely because of misuse or protection of users' data.

src/ZODB/scripts/repozo.py
mgedmin: I was torn about it, and therefore didn't suggest it.  The partial file may be useful for trying to investigate/fix the corruption (although I'm not sure how).  But the risk of the user accidentally using a truncated Data.fs is real.

Perhaps it would be best add a `.tmp` suffix to the output file, and rename it to the final name only if everything's fine.

src/ZODB/scripts/repozo.py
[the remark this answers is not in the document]
mgedmin: Yes, a base class for the three custom error classes, with one `except ... as e: sys.exit(str(e))` clause to catch them all in the main function.

I would suggest a name if I could come up with one ;)

Error?  Failure?

src/ZODB/scripts/repozo.py
vpelletier: I like the while-restoring-extension idea. Just to throw in a few more suffix ideas: `.part` ? `.incomplete` ? I am not sure how strong the dot-three extension format rule is these days.

icemac:
@Sebatyne Thank you for your PR. The Zope Foundation policy requires to sign a Committer agreement before a substantial pull request can be merged to any of its repositories. Please see 
[Zope_Foundation_Committer_Agreement](https://github.com/zopefoundation/developer_docs/raw/master/source/Zope_Foundation_Committer_Agreement-20181008.pdf) for details.

mgedmin:
Wow appveyor is deeply unhappy, but the failing test (test_via_monte_carlo) seems to be unrelated to this PR.

If the same test fails on master, I propose we ignore the appveyor failure.

mgedmin:
(Rebases/forced pushes make it impossible for me to review just the changes made since I last reviewed the code.  I'm not feeling up to re-reviewing the entire changeset right at the moment.  Please ping me if you think such a review is still needed.)

Sebatyne:
Hi,

Sorry for having paused this MR for so long, other things kicked in...

I reviewed one more time the changes today. I've also filled and email the Zope Contributor Agreement.

About Appveyor, I don't have any knowledge about this tool, but it seems a CI suite for windows ? The failing test seems to be caused a file system limitation that I don't have on linux : "[WinError 183] Cannot create a file when that file already exists". As I don't have access to a windows machine, I'll use this PR directly to run my fix attempts on Appveyor.

Regards,

Nicolas

Sebatyne:
Hi,

I'll run Appveyor on my fork of ZODB in fact, to avoid spam. I'll contact you again once all tests are OK :)

src/ZODB/scripts/repozo.py
```diff
@@ -649,12 +672,46 @@ def do_recover(options):
         log('Recovering file to stdout')
         outfp = sys.stdout
     else:
+        # Delete old ZODB before recovering backup as size of
+        # old ZODB + full partial file may be superior to free disk space
+        if os.path.exists(options.output):
+          log('Deleting old %s', options.output)
+          os.unlink(options.output)
```
mgedmin: Please use 4 space indents.

Sebatyne:
I've squashed the 2 last commits of fixup into the original commit. I'm waiting for tests to run one last time, just in case, then I'll merge.

Thanks to all for your time on this.

icemac:
@Sebatyne As you are now a member you should be able to merge this PR yourself.

PR #256, opened 2019-01-29
Spellcheck error in ZODB2.rst

sblondon:
Voalitile -> Volatile


tseaver:
@sblondon Thanks for the patch!

PR #269, opened 2019-05-14
TransactionalUndo closes storages it opens.

### LINKED ISSUE #268: DB.undo() leaks MVCC storages
I see this [in the relstorage tests](https://github.com/zodb/relstorage/issues/225). That means the underlying database connections leak.

Here's where a storage is allocated by calling `new_instance()`:

```Python traceback
  File "//relstorage/src/relstorage/tests/blob/testblob.py", line 146, in testUndoWithoutPreviousVersion
    database.undo(database.undoLog(0, 1)[0]['id'])
  File "//lib/python3.7/site-packages/ZODB/DB.py", line 997, in undo
    self.undoMultiple([id], txn)
  File "//lib/python3.7/site-packages/ZODB/DB.py", line 978, in undoMultiple
    txn.join(TransactionalUndo(self, ids))
  File "//lib/python3.7/site-packages/ZODB/DB.py", line 1060, in __init__
    db._mvcc_storage, 'undo_instance', db._mvcc_storage.new_instance)()
```

The `TransactionalUndo` object never closes the new storage.

RelStorage could potentially work around this (e.g., implement a 'undo_instance' method that returns a special storage wrapper that closes itself in `tpc_finish` and `tpc_abort`), but it seems like the more general solution would be for `TransactionalUndo` to close the storage it opened in its implementation of those methods.

jamadden:
Fixes #268.

Additionally, it no longer eagerly opens the new instance, waiting until it actually needs it, and disposes of it (not just closes it) when it is done, thus breaking the reference cycle outlined in #268.

I did the closing in both `tpc_finish` and `tpc_abort` because one or the other of those (logically) *must* be called once `tpc_begin` was called. In the implementation of `transaction`, it looks like the plain `abort()` method is also always called, but the docs don't make it clear that that's required (in fact, they say that `abort` "may also be called").

### DIFF
--- a/src/ZODB/DB.py
+++ b/src/ZODB/DB.py
@@ -14,32 +14,29 @@
 """Database objects
 """
 from __future__ import print_function
+
 import sys
 import logging
 import datetime
 import time
 import warnings
 
-from . import utils
-
-from ZODB.broken import find_global
-from ZODB.utils import z64
-from ZODB.Connection import Connection, TransactionMetaData, noop
-from ZODB._compat import Pickler, _protocol, BytesIO
-import ZODB.serialize
-
-import transaction.weakset
-
-from zope.interface import implementer
-from ZODB.interfaces import IDatabase
-from ZODB.interfaces import IMVCCStorage
-
-import transaction
-
 from persistent.TimeStamp import TimeStamp
 import six
 
-from . import POSException, valuedoc
+import transaction
+import transaction.weakset
+
+from zope.interface import implementer
+
+from ZODB import utils
+from ZODB.interfaces import IDatabase
+from ZODB.interfaces import IMVCCStorage
+from ZODB.broken import find_global
+from ZODB.utils import z64
+from ZODB.Connection import Connection, TransactionMetaData, noop
+import ZODB.serialize
+from ZODB import valuedoc
 
 logger = logging.getLogger('ZODB.DB')
 
@@ -1056,19 +1053,43 @@ class TransactionalUndo(object):
 
     def __init__(self, db, tids):
         self._db = db
-        self._storage = getattr(
-            db._mvcc_storage, 'undo_instance', db._mvcc_storage.new_instance)()
         self._tids = tids
+        self._storage = None
 
     def abort(self, transaction):
         pass
 
+    def close(self):
+        if self._storage is not None:
+            # We actually want to release the storage we've created,
+            # not close it. releasing it frees external resources
+            # dedicated to this instance, closing might make permanent
+            # changes that affect other instances.
+            self._storage.release()
+            self._storage = None
+
     def tpc_begin(self, transaction):
+        assert self._storage is None, "Already in an active transaction"
+
         tdata = TransactionMetaData(
             transaction.user,
             transaction.description,
             transaction.extension)
         transaction.set_data(self, tdata)
+        # `undo_instance` is not part of any IStorage interface;
+        # it is defined in our MVCCAdapter. Regardless, we're opening
+        # a new storage instance, and so we must close it to be sure
+        # to reclaim resources in a timely manner.
+        #
+        # Once the tpc_begin method has been called, the transaction manager will
+        # guarantee to call either `tpc_finish` or `tpc_abort`, so those are the only
+        # methods we need to be concerned about calling close() from.
+        db_mvcc_storage = self._db._mvcc_storage
+        self._storage = getattr(
+            db_mvcc_storage,
+            'undo_instance',
+            db_mvcc_storage.new_instance)()
+
         self._storage.tpc_begin(tdata)
 
     def commit(self, transaction):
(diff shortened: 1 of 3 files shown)


jamadden:
Any interested reviewers?

src/ZODB/DB.py
```diff
@@ -1056,19 +1058,38 @@ class TransactionalUndo(object):
...
     def abort(self, transaction):
         pass
 
+    def close(self):
+        if self._storage is not None:
+            self._storage.close()
+            self._storage = None
+
     def tpc_begin(self, transaction):
+        assert self._storage is None, "Already in an active transaction"
```
ale-rt: Should we raise a proper exception here?

jamadden: I think a failure of that condition would mean that the `transaction` library/protocol is broken. So to test it we'd have to use this object in a very unnatural way, violating its contract. ZODB contains numerous `assert` statements for other conditions that "can't" happen. I didn't see the value in making this one different and writing the corresponding tests.

jamadden:
I realized we should really be `release()` the new storage, not closing it. For the MVCC adapters, this doesn't make a difference (they're both no-ops), but on RelStorage there is a difference.

It would be nice to get this merged and the rest of the changes in master released.

jmuchemb:
Are you sure relstorage needs things to be done when allocating/freeing instances for undos ? I mean: relstorage does not implement undo_instance and mvccstorage does not nothing for undo instances.

If relstorage could implement undo_instance() in such a way calling release would not be needed, the ZODB code would be simpler.

jamadden:
> Are you sure relstorage needs things to be done when allocating/freeing instances for undos ? I mean: relstorage does not implement undo_instance and mvccstorage does not nothing for undo instances.

Yes. RelStorage implements `new_instance`, which this code calls (RelStorage directly implements `IMVCCStorage` so it's not ever wrapped in an mvccadapter). That allocates external resources, and those need to be closed in a timely fashion without relying on the garbage collector. Without this change, database connections leak.

> If relstorage could implement undo_instance() in such a way calling release would not be needed, the ZODB code would be simpler.

I suppose that's true, but I'm not particularly a fan of that for a few different reasons:

1) It seems like it's just the Right Thing To Do to `release()` the storage you allocated by calling `new_instance()`. The responsible party is the one that opened it, they should be the one to close it.
2) Implementing `undo_instance` in that way would require assuming things about how it is used (e.g., always closing everything at the end of `tpc_finish` and `tpc_abort`...just like the code here does, but the code here *knows* how it's using the storage).
3) `undo_instance` is not a documented part of any interface, and requiring storages to implement that would need for it to be documented, and then it would push the complexity of knowing/assuming how the storage is used out to all the storage implementations instead of keeping it centralized here.

src/ZODB/DB.py
```diff
@@ -1081,15 +1106,23 @@ def tpc_vote(self, transaction):
         self._storage.tpc_vote(transaction)
 
     def tpc_finish(self, transaction):
-        transaction = transaction.data(self)
-        self._storage.tpc_finish(transaction)
+        with contextlib.closing(self):
```
jmuchemb: The only case where `contextlib.closing` is nice is when opening several objects in a row, which means one have a single `with` statement and single indentation, rather than nested try...finally.

Here, it's abuse of sugar. The object is already open so you don't even save the line for opening. The result is slower code for questionable benefit about readability.

On the other side, the next 2 lines can be merged.

src/ZODB/DB.py
```diff
@@ -1056,19 +1058,42 @@ class TransactionalUndo(object):
...
+        # a new storage instance, and so we must close it to be sure
+        # to reclaim resources in a timely manner.
+        #
+        # Once the tpc_begin method has been called, the transaction manager will
+        # guarantee to call either `tpc_finish` or `tpc_abort`, so those are the only
+        # methods we need to be concerned about calling close() from.
+        self._storage = getattr(
+            self._db._mvcc_storage,
+            'undo_instance',
+            self._db._mvcc_storage.new_instance)()
```
jmuchemb: I prefer that `self._db._mvcc_storage` is fetched once:
```py
        storage = self._db._mvcc_storage
        self._storage = getattr(
            storage, 'undo_instance', storage.new_instance)()
```
(no preference for code style inside getattr)

src/ZODB/DB.py
jamadden:
```python
with closing(thing):
    do_stuff()
```

is still two lines shorter than 

```python
try:
    do_stuff()
finally:
    thing.close()
```

It's arguably more readable, but that's clearly a matter of opinion. `contextlib` is somewhat slower, but that's unlikely to matter here. 

There are a couple semantic differences, too, the most significant of which is that [a `with` block is safer from asynchronous exceptions](https://vorpus.org/blog/control-c-handling-in-python-and-trio/#id4) such as `KeyboardInterrupt` than a `try/finally` pair. That's *also* probably unlikely to matter much here, though.

I don't much care so I went ahead and changed it.

jmuchemb: Thanks.

> There are a couple semantic differences, too, the most significant of which is that [a `with` block is safer from asynchronous exceptions](https://vorpus.org/blog/control-c-handling-in-python-and-trio/#id4) such as `KeyboardInterrupt` than a `try/finally` pair. That's _also_ probably unlikely to matter much here, though.

The link is interesting. I already knew the race about try...except, but not that `with` can be safer. If I read correctly, it's only safer if the work to undo is done in `__enter__` (not the case of `closing`), and if `__enter__` is written in C (again not the case of `closing`).

try...finally is not less safe here.

PR #278, opened 2019-08-31
PersistentReferences are hashable

azmeuk:
In some `_p_resolveConflict` implementation on custom types, it would greatly help if I could put some values in a set. This pull request implements `PersistentReference.__hash__` and solves this problem.


jamadden:
[Equal objects](https://hynek.me/articles/hashes-and-equality/) should [have equal hashes](https://dzone.com/articles/working-with-hashcode-and-equals-in-java). This class defines what it means for instances to be equal in terms of their database and OID, but the `__hash__` implementation does not respect that. If `__hash__` is going to be implemented, I believe it needs to be implemented to match `__cmp__` and use that same information.

azmeuk:
Thank you for your review @jamadden and for the interesting links. I have updated the `__hash__` function so it takes in account the same criterias as `__cmp__` does.

azmeuk:
There are now :)

CHANGES.rst
```diff
@@ -13,12 +13,15 @@
...
+- Make ``PersistentReference`` hashable (`#278
+  <https://github.com/zopefoundation/ZODB/issues/278>`_).
+
 5.5.1 (2018-10-25)
 ==================
 
 - Fix KeyError on releasing resources of a Connection when closing the DB.
   This requires at least version 2.4 of the `transaction` package.
-  See `issue 208 <https://github.com/zopefoundation/ZODB/issues/208>`.
+  See `issue 208 <https://github.com/zopefoundation/ZODB/issues/208>`_.
```
mgedmin: Thank you!

src/ZODB/ConflictResolution.py
```diff
@@ -196,6 +196,9 @@ def __le__(self, other):
     def __ge__(self, other):
         return self.__cmp__(other) >= 0
 
+    def __hash__(self):
+        return hash((self.oid, self.database_name, self.weak))
```
mgedmin: I'm not convinced using PersistentReferences as keys in dicts/sets will work correctly when `__eq__` either returns True or raises a ValueError, but never returns False.

I guess I'll see if the tests will manage to convince me.

src/ZODB/tests/ConflictResolution.py
```diff
@@ -131,6 +138,27 @@ def checkBuggyResolve2(self):
...
+        c1 = db.open(t1)
+        c2 = db.open(t2)
+
+        c1.root()['z'][1] = "foo"
+        c2.root()['z'][1] = "bar"
+
+        t1.commit()
+        t2.commit()
+
+
```
mgedmin: I'm still not convinced.

I'd like to see what happens when you end up with two different PersistentReferences that hash() to the same value but are not equal.  I'd like to see a dict/set handle that without letting that pernicious ValueError escape.

azmeuk: Are you sure it is even possible? The hash depends on the oid (and the database name), so two objects can have the same hash only if they have the same oid (and the same database name).
In my understanding, this is not possible that two PersistentReference can have the same hash without being equal.

src/ZODB/ConflictResolution.py
azmeuk: Well. `ValueError` during a `_p_resolveConflict` end in a `ConflictError`, so I suppose I can help in some cases (when comparison is True), but it cannot be worse that currently.

src/ZODB/tests/ConflictResolution.py
mgedmin: > Are you sure it is even possible?

Theoretically, yes -- the whole point of hashing is to map a large set of values to a smaller one, and therefore hash collisions are inevitable.

In practice, maybe?  I'm not sure how tuple hashes work, but I wouldn't be surprised to find `hash((oid, '', False)) == hash((oid + constant, '', True))` for some constant value.  (I'm abusing the `+` operator here to mean arithmetic addition despite oids being constant-length bytestrings.)

azmeuk: A quick search on hash collision probability convinced me that it is safe to assume any hash is unique, whatever it hashes.

mgedmin: > A quick search on hash collision probability convinced me that it is safe to assume any hash is unique, whatever it hashes.

That is true for cryptographic hashes (SHA-256), somewhat less true for broken old cryptographic hashes (MD5 and SHA-1), and not true for plain old hashes such as Python's hash().

For example

```
>>> hash((-1, '', False)) == hash((-2, '', False))
True
```

which is not directly applicable here (OIDs are 8-byte bytestrings, not negative integers), but should be illustrative.

azmeuk: `-1` seems to be a special case for python hash.
I am not really sure what to do to convince you now :) I cannot find a testcase where two PersistentReferences could be equal without having the same hash. Do you have an idea?

mgedmin: We can always say "this is vanishingly unlikely" and let the users tell us if such collisions actually exist, by filing bugs if they get that ValueError.

mgedmin: (I just don't want to be responsible for making that decision, which is why I'm declining to approve or reject this PR.  Luckily I'm not the only maintainer of ZODB!  Actually not even one of the main maintainers.)

azmeuk: > We can always say "this is vanishingly unlikely" and let the users tell us if such collisions actually exist, by filing bugs if they get that ValueError.

This is how I feel :) Ascendant compatibility is assured.
Who can I summon to review this?

mgedmin: `git shortlog -s|sort -n|tail` tells me the people with most commits (who probably know most about ZODB) are

```
    59	Shane Hathaway
    87	Jason Madden
    88	Marius Gedminas
   158	Barry Warsaw
   166	Tres Seaver
   281	Fred Drake
   318	Guido van Rossum
   595	Tim Peters
   899	Jeremy Hylton
  1591	Jim Fulton
```

(I've no idea how I got on that list!  Fixing tests/warnings?)

icemac:
I do not feel competent to review this PR.

src/ZODB/ConflictResolution.py
tseaver: @azmeuk I think @mgedmin is correct here:  without reliable equality semantics, hashability cannot be useful.

azmeuk: Well I find it useful and I would love this to come into ZODB so I could use it in my projects :)
@tseaver would you agree with those points?

- There is no backward compatibility;
- There is no danger for existing data;
- It might help in some cases;
- It might have unknown behavior in some other cases.

My suggestion is to merge, and let the users open tickets if they find unwanted behavior, as mgedmin suggested it.

jamadden: I think I'm with @tseaver and @mgedmin here and it's probably not a good idea for ZODB to introduce an attractive footgun into conflict resolution.

The claim is that it's "not possible to compare different weak references", so there's not a total order defined and you can't put these in a BTree based Set either.  But you could violate that if you chose, either by making an artificial key to wrap the reference, or by storing the references in a dictionary keyed on their components. You'd be comparing different references, which is apparently not generally reliable, but perhaps in your special circumstance that's good enough?


```python
s = BTrees.OOBTree.TreeSet()
s.add((ref.oid, ref.database_name, ref.weak, ref))

d = {}
d[(ref.oid, ref.database_name, ref.weak)] = ref
```

jmuchemb:
If I understand correctly, the current `__cmp__` & other are implemented as if we wanted the same result as outside conflict resolution, which is most of the time impossible.

@azmeuk You should first describe your use case and how you'd like to compare your persistent references.

> If `__hash__` is going to be implemented, I believe it needs to be implemented to match `__cmp__` and use that same information.

No. The following implementation is always valid:
```py
def __hash__(self):
    return 0
```
But obviously inefficient.

jamadden:
> No. The following implementation is always valid:
> def __hash__(self):
>    return 0
> But obviously inefficient.

True enough, because it preserves the invariant that equal objects have equal hashcodes. The original implementation here, `return id(self)` does not preserve that invariant: equal objects no longer had equal hash codes.  That's exactly why Python actually *deletes* the default `__hash__` method when you define `__eq__`:

```pycon
>>> class OverrideNothing:
...     pass
...
>>> set([OverrideNothing(), OverrideNothing()])
{<__main__.OverrideNothing object at 0x107c58cc0>,
 <__main__.OverrideNothing object at 0x107c88720>}
>>> hash(OverrideNothing())
276584670
>>> class OverrideEq:
...     def __eq__(self, other):
...         return self is other
...
>>> set([OverrideEq(), OverrideEq()])
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'OverrideEq'
>>> hash(OverrideEq())
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'OverrideEq'
```

Restoring it, as was first proposed, isn't useful. So in the context of an object such as this one that defines a notion of equality, either `__hash__` is implemented to use (some subset of) the same information that equality uses, or it returns a constant. Or as [the documentation puts it](https://docs.python.org/3/reference/datamodel.html#object.__hash__):

> The only required property is that objects which compare equal have the same hash value; it is advised to mix together the hash values of the components of the object that also play a part in comparison of objects by packing them into a tuple and hashing the tuple. Example:
```python
def __hash__(self):
    return hash((self.name, self.nick, self.color))
```

Krilivye:
Hi there,
We're interested in this fix too.

For the context, we need that for a three-way dictionary deep merge in conflict resolution

jamadden:
I remain opposed to this implementation because it's unsafe. Whether it appears to work or not depends on both (1) whether there are any hash conflicts and (2) the exact details of the implementation of the data structure using it (bucket sizes and distribution, probing strategy; basically, how likely it is to invoke `__eq__` even if there are no hash conflicts).

We can't control (1), and it can change from run to run given hash seed randomization. Sometimes `(0, 'db1', True)` might hash differently then `(0, 'db2', True)` and sometimes it might hash the same.

We can't control (2), and it might change on different versions of Python or different implementations.

d-maurer:
Jason Madden wrote at 2020-9-18 10:17 +0000:
>I remain opposed to this implementation because it's unsafe. Whether it appears to work or not depends on both (1) whether there are any hash conflicts and (2) the exact details of the implementation of the data structure using it (bucket sizes and distribution, probing strategy; basically, how likely it is to invoke `__eq__` even if there are no hash conflicts).

Why not: two persistent references are equal iff they reference
the same object (i.e. same database and same _p_oid)?
This looks quite safe for me.
The hash can then be derived from database and _p_oid.

jamadden:
The reasons described above are sound as far as they go, but here's the most fundamental reason: a `PersistentReference` *isn't* the underlying object, and attempts to treat it like so during conflict resolution are likely to lead to problems due to differences between how `PersistentReference` hashes (and compares) and how the underlying object hashes and compares.

Suppose you have this object:
```python
class O(Persistent):
    def __init__(self, value):
        self.value = value
    def __hash__(self): return hash(self.value)
    def __eq__(self, other): return self.value == other.value
```

Now, there are some relationships that hold:
```python
O('abc') == O('abc')
set([O('abc')]) == set([O('abc')])
```

Even when those objects are added to a database and get distinct OIDs, those relationships hold.

However, once conflict resolution kicks in, and those objects are replaced by `PersistentReference`, *those relationships no longer hold*.  The custom code that defines them doesn't get a chance to run.

Currently, `PersistentReference` can say that it is equal to another `PersistentReference` exactly and only when it represents the same object. For any other case, it refuses to guess. 

Likewise, it refuses to hash, because it has no idea how the underlying object chooses to has, and anything else can produce incorrect results. During conflict resolution, object `O('abc')`  in one state will have OID 1, let's say, and in the other state it will have OID 2. Currently those fail to hash, but if we add a hash, then it will seem like they are *not* the same object.
```python
set([PersistentReference(O('abc') with oid 1)]) != set([PersistentReference(O('abc') with oid 2)])
```
The relationship is broken and conflict resolution becomes inconsistent (or much harder).

The safest thing is to refuse to guess, as `__eq__` does.

[BTrees documents this in a round-about way](https://btrees.readthedocs.io/en/latest/overview.html#total-ordering-and-persistence), though it still refers to Python 2 semantics.

This was discovered before: Raising `ValueError` for non-identical comparisons was [added in ZODB 3.8](https://github.com/zopefoundation/ZODB/commit/520c44bde2cfab2e2c55cfb5c683c7cf07fe9565#diff-1cad7780efa40f5e02a4280497e8ad4dR309) to address "serious problems" leading to inconsistencies in BTrees. The same applies for hashed-based data structures. 

Why has that not been a "serious problem" thus far? Hash-based persistent data structures, of which the main one I'm aware of is `PersistentMapping,` don't participate in conflict resolution. If it did, then under Python 3 the automatic `TypeError` we currently get would generally prevent the problem (it's still an issue on Python 2, though, so I would support explicitly adding a `__hash__` that raises `ValueError` on Python 2, just to be safe).

d-maurer:
Jason Madden wrote at 2020-9-18 10:39 -0700:
>The reasons described above are sound as far as they go, but here's the most fundamental reason: a `PersistentReference` *isn't* the underlying object, and attempts to treat it like so during conflict resolution are likely to lead to problems due to differences between how `PersistentReference` hashes (and compares) and how the underlying object hashes and compares.

I fully agree with you.

Krilivye:
Big thanks for the insight @jamadden.

PR #280, opened 2019-09-17
Use weakref.WeakSet rather than transaction.weakset.WeakSet

jmuchemb:


src/ZODB/DB.py
```diff
@@ -299,20 +297,6 @@ def clear(self):
...
-        for pool in six.itervalues(self.pools):
-            result.update(pool.all)
-        return frozenset(result)
-
-    @property
-    def test_available(self):
-        result = []
-        for pool in six.itervalues(self.pools):
-            result.extend(pool.available)
-        return tuple(result)
```
jmuchemb: if you wonder, it looks like dead code. I found no use in this repository or ZEO, and the commit message that added these methods doesn't explain them.

PR #281, opened 2019-09-21
Tests: Always pass bytes for the previous serial param of storage.store

jamadden:
In a few cases, they were passing native strings, which complicates things on Python 3.

This is consistent with what persistent.Persistent and Connection actually do.

RelStorage previously had a special case just for these tests.

### DIFF
--- a/src/ZODB/scripts/migrate.py
+++ b/src/ZODB/scripts/migrate.py
@@ -80,12 +80,12 @@ import getopt
 import marshal
 import profile
 
+from persistent.timestamp import TimeStamp
 from ZODB import utils
-from ZODB import StorageTypes
-from ZODB.TimeStamp import TimeStamp
+from ZODB import StorageTypes # XXX: This import does not exist
+
 
 PROGRAM = sys.argv[0]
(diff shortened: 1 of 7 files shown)


src/ZODB/scripts/migrate.py
```diff
@@ -85,7 +85,7 @@
 from ZODB.TimeStamp import TimeStamp
 
 PROGRAM = sys.argv[0]
-ZERO = '\0'*8
+ZERO = b'\0' * 8
```
jmuchemb:
```py
ZERO = utils.z64
```

jmuchemb: or replace the only use of `ZERO` with `utils.z64`

jamadden: Thanks for the suggestions! I was trying to keep the changes as minimal as possible, but since you brought it up I'm more than happy to DRY.

jamadden: This script doesn't even run or import because there  is no `from ZODB import StorageTypes`

PR #293, opened 2019-12-30
fix #292 improve volatile attribute documentation

### LINKED ISSUE #292: Better documentation of _v_*
In https://github.com/zopefoundation/ZODB/blob/master/doc/guide/writing-persistent-objects.rst#special-attributes The `_v_` attribute is documented. 

I never been sure how it works exactly. Now I had the need to look deeper into it. I always thought (for >15years), in a ZEO setup `_v_` is local to the connection and does not get propagated to other clients, which is *wrong*. No idea where I got that from... but in fact the `_v_` is propagated to the other connections/clients after commit. It is just not stored in the Data.fs as far as I understand.

So, lucky me, this made it simple for my case.

But the documentation is bit thin here. Maybe this needs to go in the ZEO documentation, because its ZEO special? How does RelStorage and other backends do here?

jensens:
I took the statements and insight from #292 and enhanced the documentation.

### DIFF
--- a/doc/guide/writing-persistent-objects.rst
+++ b/doc/guide/writing-persistent-objects.rst
@@ -272,20 +272,13 @@ the value ``None`` rather than a dictionary for :ref:`ghosts
 Volatile Attributes
 -------------------
 
-Attributes with names starting with ``_v_`` are treated as volatile.
-They are useful for caching data that can be computed from saved data
-and should not be saved [#cache]_.
+Attributes with names starting with ``_v_`` are volatile, 
+they are never serialized and not saved to the database.
+They are useful for caching data that can be computed from other data[#cache]_.
 
-They are never serialized and not saved to the database.
 Volatile attributes are local to a specific active object in memory and
-thus to a specific connection.
-They should be treated as though they can disappear between transactions:
-If an object is removed from the connection cache the volatile attribute is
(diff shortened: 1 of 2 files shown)


doc/guide/writing-persistent-objects.rst
```diff
@@ -264,18 +264,31 @@ _p_jar
...
+Attributes with names starting with ``_v_`` are treated as volatile.
+They are useful for caching data that can be computed from saved data
+and should not be saved [#cache]_.
+
+They are never serialized and not saved to the database.
+Volatile attributes are local to a specific active object in memory and
+thus to a specific connection.
+They should be treated as though they can disappear between transactions:
+If an object is removed from the connection cache the volatile attribute is
+gone.
```
jimfulton: s/gone/lost/

jensens:
@jimfulton ok, I reduced it a bit, removed the "insight" part about thread/connection share (maybe this section is the wrong place for this information) 

I think it is not as rarely used as you think. And if one uses volatile attributes, it is important to understand them without digging through its source code.

icemac:
Is this PR ready merge?

jensens:
@icemac sorry, I always forget, in `zopefoundation` it is self-merge after approval.

Issue #285, opened 2019-11-18
UnicodeDecodeError when loading pickle from ZODB that was migrated from Python 2 to 3


pbauer:
I have 18 seemingly broken `OISet` pickles in by ZODB after migrating it from 2.7 to 3.7.
In py2 all seems to be ok and I can load them from the DB. The same pickle throws a `UnicodeDecodeError` when I try to load it in Python 3.

I don't know what the actual object is or where in the tree it is used (it's from a medium-sized Plone site). The transaction-description points to `plone.app.iterate` (working-copies for content) and to content that no longer exists. So I think I could simply remove them.

I migrated the ZODB with zodbupdate: `./bin/zodbupdate --convert-py3 --file=var/filestorage/Data.fs --encoding=utf8`. I've done this multiple times so far for various projects and never had this issue.

Below is information I gathered while debugging in py2 and py3.

Python 2:

```
>>> from ZODB.utils import p64
>>> from plone import api
>>> portal = api.portal.get()
>>> app = portal.__parent__
>>> oid = 0x2c0ab6
>>> oid = p64(oid)
>>> oid
'\x00\x00\x00\x00\x00,\n\xb6'
>>> obj = app._p_jar.get(oid)
>>> obj
OISet([
 <InterfaceClass OFS.EtagSupport.EtagBaseInterface>,
 <InterfaceClass Acquisition.interfaces.IAcquirer>,
 <SchemaClass plone.app.dexterity.behaviors.discussion.IAllowDiscussion>,
 <InterfaceClass zope.annotation.interfaces.IAnnotatable>,
 <InterfaceClass zope.annotation.interfaces.IAttributeAnnotatable>,
 <InterfaceClass plone.uuid.interfaces.IAttributeUUID>,
 <InterfaceClass Products.CMFDynamicViewFTI.interfaces.IBrowserDefault>,
 <InterfaceClass Products.CMFCore.interfaces.ICatalogAware>,
 <InterfaceClass Products.CMFCore.interfaces.ICatalogableDublinCore>,
 <InterfaceClass zope.location.interfaces.IContained>,
 <InterfaceClass Products.CMFCore.interfaces.IContentish>,
 <class 'grokcore.component.interfaces.IContext'>,
 <InterfaceClass OFS.interfaces.ICopySource>,
 <InterfaceClass webdav.interfaces.IDAVResource>,
 <InterfaceClass plone.dexterity.interfaces.IDexterityContent>,
[...]

PR #295, opened 2020-03-02
Fix unbound local error when using the fsoids.py script 

pbauer:
Fix issue mentioned in https://github.com/zopefoundation/ZODB/issues/285#issuecomment-584357646

Closes #329 (duplicate)


src/ZODB/serialize.py
```diff
@@ -699,7 +699,7 @@ def get_refs(a_pickle):
         if isinstance(reference, tuple):
             oid, klass = reference
         elif isinstance(reference, (bytes, str)):
-            data, klass = reference, None
+            oid, klass = reference, None
```
jmuchemb: Split in 2 lines to avoid the tuple packing/unpacking.

mgedmin: But this way it's nicely symmetric with the case above, where reference is a tuple.

mgedmin:
Python 2.7 fails on Appveyor with syntax errors in mock.

I'm pretty sure this is because Appveyor's default pip is too old to pay attention to python_requires metadata and installs the wrong version of mock.

A `python -m pip install -U pip` step in appveyor.yml ought to fix things.  (Maybe it should be in a separate PR, merged before this one?)

mgedmin:
> A python -m pip install -U pip step in appveyor.yml ought to fix things. (Maybe it should be in a separate PR, merged before this one?)

I should've looked at the appveyor.yml before making assumptions.  We already upgrade pip to the latest version.

mock is actually installed by buildout.  Suggestion: instead of buildout let's use `tox -e py` in the appveyor.yml.

ale-rt:
I would squash and merge the commits if the build passes

icemac:
I restarted the PyPy3 job which had failed. But now it is green. @pbauer Feel free to merge.

mgedmin:
Oops my finger slipped and I accidentally did the suggested squash-and-merge!

Thank you everyone!

PR #296, opened 2020-03-05
Fix test compatibility with persistent 4.6; test CPython with PURE_PYTHON

jamadden:


mgedmin:
Appveyor on Python 2.7 is unhappy because it uses zc.buildout to install stuff, and zc.buildout doesn't pay attention to python_requires and gets a version of mock that doesn't work on 2.7.

jamadden:
Thanks! I'll take a look at appveyor in a separate branch.

PR #298, opened 2020-03-13
FileStorage: Save committed transaction to disk even if changed data is empty

navytux:
ZODB tries to avoid saving empty transactions to storage on
`transaction.commit()`. The way it works is: if no objects were changed
during ongoing transaction, ZODB.Connection does not join current
TransactionManager, and transaction.commit() performs two-phase commit
protocol only on joined DataManagers. In other words if no objects were
changed, no tpc_*() methods are called at all on ZODB.Connection at
transaction.commit() time.

This way application servers like Zope/ZServer/ERP5/... can have
something as

    try:
        # process incoming request
        transaction.commit()    # processed ok
    except:
        transaction.abort()
        # problem: log + reraise

in top-level code to process requests without creating many on-disk
transactions with empty data changes just because read-only requests
were served.

Everything is working as intended.

However at storage level, FileStorage currently also checks whether
transaction that is being committed also comes with empty data changes,
and _skips_ saving transaction into disk *at all* for such cases, even
if it has been explicitly told to commit the transaction via two-phase
commit protocol calls done at storage level.

This creates the situation, where contrary to promise in
ZODB/interfaces.py(*), after successful tpc_begin/tpc_vote/tpc_finish()
calls made at storage level, transaction is _not_ made permanent,
despite tid of "committed" transaction being returned to caller. In other
words FileStorage, when asked to commit a transaction, even if one with
empty data changes, reports "ok" and gives transaction ID to the caller,
without creating corresponding transaction record on disk.

This behaviour is

a) redundant to application-level avoidance to create empty transaction
   on storage described in the beginning, and

b) creates problems:

The first problem is that application that works at storage-level might
be interested in persisting transaction, even with empty changes to
data, just because it wants to save the metadata similarly to e.g.
`git commit --allow-empty`.

The other problem is that an application view and data in database
become inconsistent: an application is told that a transaction was
created with corresponding transaction ID, but if the storage is
actually inspected, e.g. by iteration, the transaction is not there.
This, in particular, can create problems if TID of committed transaction
is reported elsewhere and that second database client does not find the
transaction it was told should exist.

I hit this particular problem with wendelin.core. In wendelin.core,
there is custom virtual memory layer that keeps memory in sync with
data in ZODB. At commit time, the memory is inspected for being dirtied,
and if a page was changed, virtual memory layer joins current
transaction _and_ forces corresponding ZODB.Connection - via which it
will be saving data into ZODB objects - to join the transaction too,
because it would be too late to join ZODB.Connection after 2PC process
has begun(+). One of the format in which data are saved tries to
optimize disk space usage, and it actually might happen, that even if
data in RAM were dirtied, the data itself stayed the same and so nothing
should be saved into ZODB. However ZODB.Connection is already joined
into transaction and it is hard not to join it because joining a
DataManager when the 2PC is already ongoing does not work.

This used to work ok with wendelin.core 1, but with wendelin.core 2 -
where separate virtual filesystem is also connected to the database to
provide base layer for arrays mappings - this creates problem, because
when wcfs (the filesystem) is told to synchronize to view the database
@tid of committed transaction, it can wait forever waiting for that, or
later, transaction to appear on disk in the database, creating
application-level deadlock.

I agree that some more effort might be made at wendelin.core side to
avoid committing transactions with empty data at storage level.

However the most clean way to fix this problem in my view is to fix
FileStorage itself, because if at storage level it was asked to commit
something, it should not silently skip doing so and dropping even non-empty
metadata + returning ok and committed transaction ID to the caller.

As described in the beginning this should not create problems for
application-level ZODB users, while at storage-level the implementation
is now consistently matching interface and common sense.

----

(*) tpc_finish: Finish the transaction, making any transaction changes permanent.
    Changes must be made permanent at this point.
    ...

    https://github.com/zopefoundation/ZODB/blob/5.5.1-35-gb5895a5c2/src/ZODB/interfaces.py#L828-L831

(+) https://lab.nexedi.com/kirr/wendelin.core/blob/9ff5ed32/bigfile/file_zodb.py#L788-822

### DIFF
--- a/src/ZODB/tests/testFileStorage.py
+++ b/src/ZODB/tests/testFileStorage.py
@@ -326,6 +326,35 @@ class FileStorageTests(
             self._storage._files.flush = lambda: None
             self.checkFlushAfterTruncate(True)
 
+    def checkCommitWithEmptyData(self):
+        """
+        Verify that transaction is persisted even if it has no data, or even
+        both no data and empty metadata.
+        """
+
+        # verify:
+        # - commit with empty data but non-empty metadata
+        # - commit with empty data and empty metadata
+        #   (the fact of commit carries information by itself)
+        stor = self._storage
+        for description in (u'commit with empty data', u''):
+            t = TransactionMetaData(description=description)
+            stor.tpc_begin(t)
+            stor.tpc_vote(t)
+            head = stor.tpc_finish(t)
+            self.assertEqual(head, stor.lastTransaction())
+
+            v = list( stor.iterator(start=head, stop=head) )
+            self.assertEqual(len(v), 1)
+            trec = v[0] # FileStorage.TransactionRecord or hexstorage.Transaction
+            self.assertEqual(trec.tid, head)
+            self.assertEqual(trec.user,          b'')
+            self.assertEqual(trec.description,   description.encode('utf-8'))
+            self.assertEqual(trec.extension,     {})
+            drecv = list(trec)
+            self.assertEqual(drecv, [])
+
+
 class FileStorageHexTests(FileStorageTests):
 
     def open(self, **kwargs):
--- a/src/ZODB/FileStorage/FileStorage.py
+++ b/src/ZODB/FileStorage/FileStorage.py
@@ -782,8 +782,6 @@ class FileStorage(
                 raise StorageTransactionError(
                     "tpc_vote called with wrong transaction")
             dlen = self._tfile.tell()
-            if not dlen:
-                return # No data in this trans
             self._tfile.seek(0)
             user, descr, ext = self._ude
 
@@ -828,22 +826,19 @@ class FileStorage(
         return tid
 
     def _finish(self, tid, u, d, e):
-        # If self._nextpos is 0, then the transaction didn't write any
-        # data, so we don't bother writing anything to the file.
-        if self._nextpos:
-            # Clear the checkpoint flag
-            self._file.seek(self._pos+16)
-            self._file.write(as_bytes(self._tstatus))
-            try:
-                # At this point, we may have committed the data to disk.
-                # If we fail from here, we're in bad shape.
-                self._finish_finish(tid)
-            except:
-                # Ouch.  This is bad.  Let's try to get back to where we were
-                # and then roll over and die
-                logger.critical("Failure in _finish. Closing.", exc_info=True)
-                self.close()
-                raise
+        # Clear the checkpoint flag
+        self._file.seek(self._pos+16)
+        self._file.write(as_bytes(self._tstatus))
+        try:
+            # At this point, we may have committed the data to disk.
+            # If we fail from here, we're in bad shape.
+            self._finish_finish(tid)
+        except:
+            # Ouch.  This is bad.  Let's try to get back to where we were
+            # and then roll over and die
+            logger.critical("Failure in _finish. Closing.", exc_info=True)
+            self.close()
+            raise
 
     def _finish_finish(self, tid):
         # This is a separate method to allow tests to replace it with


navytux:
Updated the patch with the following interdiff to fix Python3 failures:

```diff
--- a/src/ZODB/tests/testFileStorage.py
+++ b/src/ZODB/tests/testFileStorage.py
@@ -348,8 +348,8 @@ def checkCommitWithEmptyData(self):
             self.assertEqual(len(v), 1)
             trec = v[0] # FileStorage.TransactionRecord or hexstorage.Transaction
             self.assertEqual(trec.tid, head)
-            self.assertEqual(trec.user,          u'')
-            self.assertEqual(trec.description,   description)
+            self.assertEqual(trec.user,          b'')
+            self.assertEqual(trec.description,   description.encode('utf-8'))
             self.assertEqual(trec.extension,     {})
             drecv = list(trec)
             self.assertEqual(drecv, [])
```

navytux:
CI: appveyor is all green, travis: mostly green with 2 strange [SyntaxError](https://travis-ci.org/github/zopefoundation/ZODB/jobs/662040474?utm_medium=notification&utm_source=github_status) [failures](https://travis-ci.org/github/zopefoundation/ZODB/jobs/662040479?utm) in pygments egg that look to be completely unrelated to my patch.

mgedmin:
The failures are caused by the travis build relying on zc.buildout, which doesn't support python_requires metadata and installs Python-3-only packages on Python 2.  This is a problem unrelated to your PR, and I think we should fix it first on master before accepting additional PRs.

navytux:
@mgedmin thanks for feedback on CI status.

vpelletier:
> One of the format in which data are saved tries to
> optimize disk space usage, and it actually might happen, that even if
> data in RAM were dirtied, the data itself stayed the same and so nothing
> should be saved into ZODB.

Without removing any merit from this merge request (returning a TID without it being present in-storage looks inconsistent on its own to me, and the FileStorage-level check looks redundant and good to remove), this makes me think of another API: Would it make sense to do this disk space optimization work in a [before-commit hook](https://zodb.readthedocs.io/en/latest/transactions.html#before-commit-hooks) ? I believe it is still fine to register other transaction managers at this point.

navytux:
@vpelletier, thank for feedback. This is exactly what I meant with

> I agree that some more effort might be made at wendelin.core side to
> avoid committing transactions with empty data at storage level.

wendelin.core alredy uses `beforeCompletion` synchronizer, which is [equivalent](https://github.com/zopefoundation/transaction/blob/3.0.0-3-g10cc231/transaction/_transaction.py#L267-L269) to before-commit hooks, and already joins another data manager to transaction from there [1].

The writeout process (the act of writing dirtied pages in RAM into ZODB) naturally fits into two-phase commit protocol where RAM pages are written out into ZODB at commit time [2]. The code that runs at beforeCompletion time, could indeed first go through all dirtied pages, reread their in-database state, and undirty a page if it finds out that data content of that page did not actually changed.

However usually most of the time there are changes and what is really needed is to represent δ of those changes efficiently on storage. For case of small changes we currently split 2MB block of data into 512 4K chunks each represented by separate ZODB object. This allows to save space on small incremental changes, but makes loading data from arrays much slower, because in this scheme, for every block, 512 load requests have to be issued into storage servers, and non of the production storages we currently use handle that kinds of workload simultaneously.

In other words deduplicating on client has significant cost for reads and the only way to have both 1) small storage growth on incremental updates, and 2) low latency for block reads, is to implement data deduplication on storage server side.

For this reason I was not making futher enhancements on client side with respect to dedupliction.

And as you say, even if wendelin.core behaviour is somewhat non-optimal in this-issue case, the behaviour of FileStorage to silently skip persisting transaction despite reporint to caller transaction ID as if the transaction was committed, is inconsistent and is better to fix.


[1] https://lab.nexedi.com/kirr/wendelin.core/blob/9ff5ed32/bigfile/file_zodb.py#L788-822
[2] https://lab.nexedi.com/kirr/wendelin.core/blob/9ff5ed32/bigfile/file_zodb.py#L833-888
[3] https://lab.nexedi.com/kirr/wendelin.core/blob/9ff5ed32/bigfile/file_zodb.py#L77-124

navytux:
Rebased the patch on top of master. CI is all green now.

src/ZODB/tests/testFileStorage.py
```diff
@@ -326,6 +326,35 @@ def checkFlushNeededAfterTruncate(self):
...
+        Verify that transaction is persisted even if it has no data, or even
+        both no data and empty metadata.
+        """
+
+        # verify:
+        # - commit with empty data but non-empty metadata
+        # - commit with empty data and empty metadata
+        #   (the fact of commit carries information by itself)
+        stor = self._storage
+        for description in [u'commit with empty data', u'']:
```
jmuchemb: Can you remove the square brackets ?

navytux: Do you mean change list -> tuple?

jmuchemb: Yes. But parentheses aren't required.

navytux: After being with python for ~20 years I've discovered that `for x in 1,2,3:` (i.e. without brackets) also works. I don't think this is cleanier, but if required I can indeed change to this style.

For the reference [ZODB article](https://github.com/zopefoundation/ZODB/blob/master/doc/articles/ZODB1.rst) uses the same style I used initially:

```
>>> for name in ['Mary', 'Joe', 'Bob']:
        ...
```

jmuchemb: My main concern is about the fact that lists have to be built at runtime, whereas immutable objects (here a tuple of strings) don't.

navytux: It indeed used to be so on python2, but on python3 it does not produce any difference as iterating through both constant tuples and lists produces exactly the same bytecode (using tuple inside). At the same time I find the style where iterated list/tuple is explicitly marked as such - using either `[]` or `()` to be more clear. While personally I find `[` to be more visually clear, I've changed the patch to use tuple explicitly as requested. 

```py
def f_tuple():
    for i in ('hello', 'world', 'zzz'):
        pass

def f_list():
    for i in ['hello', 'world', 'zzz']:
        pass

from dis import dis
print('\nTUPLE:')
dis(f_tuple)
print('\nLIST:')
dis(f_list)
```

```
$ python3 x.py 

TUPLE:
  2           0 SETUP_LOOP              12 (to 14)
              2 LOAD_CONST               1 (('hello', 'world', 'zzz'))
              4 GET_ITER
        >>    6 FOR_ITER                 4 (to 12)
              8 STORE_FAST               0 (i)

  3          10 JUMP_ABSOLUTE            6
        >>   12 POP_BLOCK
        >>   14 LOAD_CONST               0 (None)
             16 RETURN_VALUE

LIST:
  6           0 SETUP_LOOP              12 (to 14)
              2 LOAD_CONST               1 (('hello', 'world', 'zzz'))
              4 GET_ITER
        >>    6 FOR_ITER                 4 (to 12)
              8 STORE_FAST               0 (i)

  7          10 JUMP_ABSOLUTE            6
        >>   12 POP_BLOCK
        >>   14 LOAD_CONST               0 (None)
             16 RETURN_VALUE
```

```
kirr@deco:~/tmp/trashme$ python2 x.py 

TUPLE:
  2           0 SETUP_LOOP              14 (to 17)
              3 LOAD_CONST               4 (('hello', 'world', 'zzz'))
              6 GET_ITER            
        >>    7 FOR_ITER                 6 (to 16)
             10 STORE_FAST               0 (i)

  3          13 JUMP_ABSOLUTE            7
        >>   16 POP_BLOCK           
        >>   17 LOAD_CONST               0 (None)
             20 RETURN_VALUE        

LIST:
  6           0 SETUP_LOOP              23 (to 26)
              3 LOAD_CONST               1 ('hello')
              6 LOAD_CONST               2 ('world')
              9 LOAD_CONST               3 ('zzz')
             12 BUILD_LIST               3
             15 GET_ITER            
        >>   16 FOR_ITER                 6 (to 25)
             19 STORE_FAST               0 (i)

  7          22 JUMP_ABSOLUTE           16
        >>   25 POP_BLOCK           
        >>   26 LOAD_CONST               0 (None)
             29 RETURN_VALUE        
```

navytux:
Patch amended with the following interdiff as requested by @jmuchemb:

```diff
--- a/src/ZODB/tests/testFileStorage.py
+++ b/src/ZODB/tests/testFileStorage.py
@@ -337,7 +337,7 @@ def checkCommitWithEmptyData(self):
         # - commit with empty data and empty metadata
         #   (the fact of commit carries information by itself)
         stor = self._storage
-        for description in [u'commit with empty data', u'']:
+        for description in (u'commit with empty data', u''):
             t = TransactionMetaData(description=description)
             stor.tpc_begin(t)
             stor.tpc_vote(t)
```

navytux:
Thanks for applying the patch.

Issue #284, opened 2019-10-25
zodb.org certificate


azmeuk:
Hi,
https://www.zodb.org/ has a certificate problem.
It would be great to have https by default.

Éloi

icemac:
@jimfulton Are you able to fix the certificate of https://www.zodb.org?

jugmac00:
Afaik we had this problem at another repository also.. but cannot recall which one.

Basically the problem is that we use the wildcard-cert from readthedocs for our own server.

So I guess either we'd have to redirect completely to the readthedocs site, or get a own cert via letsencrypt and reconfigure the web server.

azmeuk:
A quick way out would be to redirect https://www.zodb.org to https://zodb-docs.readthedocs.io/en/latest/

azmeuk:
I have updated the URL in the documentation so https://www.zodb.org is now https://zodb-docs.readthedocs.io , but I cannot do it on the Github project description. Can someone with enough power do this?

mgedmin:
I can!

azmeuk:
Thank you!

PR #299, opened 2020-03-14
Some documentation love

azmeuk:
- Splitted `index.rst` in another `introduction.rst` file, because it was too loaded for a homepage.
- I moved all the `*.txt` files from the source directory to the documentation directory, renamed them to `.rst` and fixed the titles so it produce a coherent hierarchy.
- The documentation can be built with `tox -e docs`
- I switched references to www.zodb.org to https://zodb-docs.readthedocs.io as the certificate for zodb.org is not valid. (See #284)


doc/collaborations.rst
```diff
@@ -1,14 +1,13 @@
-=======================
 Collabortation Diagrams
```
jamadden: It's not clear to me that it's useful to publish this document.

doc/conf.py
```diff
@@ -41,10 +46,10 @@
 master_doc = 'index'
 
 # General information about the project.
-project = u'ZODB'
-copyright = u'2009-2016, Zope Foundation'
+project = 'ZODB'
+copyright = '2009-2020, Zope Foundation'
 
-# The version info for the project you're documenting, acts as replacement for
+# The version info for the project yo're documenting, acts as replacement for
```
jamadden: Typo.

src/ZODB/tests/testUtils.py
```diff
@@ -156,6 +156,6 @@ class ExampleClass(object):
 def test_suite():
     suite = unittest.defaultTestLoader.loadTestsFromName(__name__)
     suite.addTest(
-        doctest.DocFileSuite('../utils.txt', checker=checker)
+        doctest.DocFileSuite('../../../doc/utils.rst', checker=checker)
```
jamadden: I'm somewhat concerned that the file movements will cause problems for other projects that extend ZODB's tests. They'll potentially need access to these files if just ZODB is installed, which I don't think they'll get if they're not part of the source tree.

doc/ConflictResolution.rst
```diff
@@ -1,9 +1,8 @@
-===================
```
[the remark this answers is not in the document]
azmeuk: Done.

doc/conf.py
azmeuk: Done.

doc/collaborations.rst
azmeuk: Not sure either, and honestly, without more explanation and context I would not know what useful information to read.
I was thinking that putting all the articles on the light, even technical ones like this, could (maybe) make people want to improve it.

Still, I can unpublish it if you prefer.

src/ZODB/tests/testUtils.py
azmeuk: I can let them where they were and just include them in the documentation. Would that be better?

doc/collaborations.rst
jamadden: My major concern is that, with just a quick glance, I saw things that I'm pretty sure are now just flat-out wrong. It refers to a `register` method on arbitrary resource managers, but that was deprecated functionality and is no longer supported. There's also talk of a `discard` method that I don't think exists.

If you find value in publishing it, that's fine, but I would suggest adding a big disclaimer to the top:
```
.. caution:: 
    This document hasn't been reviewed since 2005 
    and is likely out of date.
```

src/ZODB/tests/testUtils.py
jamadden: (I know RelStorage has a tight relationship to the ZODB test suite. I haven't checked to see if moving the files would cause it any problems.)

jamadden: > I can let them where they were and just include them in the documentation. Would that be better?

I suspect that would be safer, yes.

doc/collaborations.rst
azmeuk: Ok I have put disclaimers on the old files.

src/ZODB/tests/testUtils.py
azmeuk: Done.

doc/subtransactions.rst
```diff
@@ -2,6 +2,10 @@
 Subtransactions in ZODB 3
 =========================
 
+.. caution::
```
jamadden: Good catch, this file is *definitely* out of date. Subtransactions were replaced with savepoints, and none of those function calls taking a parameter will work any more.

I think this could just be completely removed.

azmeuk: Done.

azmeuk:
@jamadden Thanks for reviewing this one. As usual it seems that the RTD doc does not build :)

jamadden:
Sorry, I have no ability to update [https://zodb-docs.readthedocs.io](https://readthedocs.org/projects/zodb-docs/). The only owner is @jimfulton

PR #301, opened 2020-03-17
Removed log.ini

azmeuk:
This file has not moved in 16 years, its name does not appear anywhere, and tests are OK without it.


azmeuk:
Indeed, the only other `.ini` in the repository is `tox.ini`.

jamadden:
Ah, I forgot about that one. So no changes needed then. Thanks.

PR #310, opened 2020-05-20
Fix inconsistent resolution order with zope.interface v5.

janjaapdriessen:


tox.ini
```diff
@@ -13,6 +13,8 @@ commands =
     zope-testrunner --test-path=src []
 deps =
     .[test]
+setenv =
+    ZOPE_INTERFACE_STRICT_IRO = 1
```
jamadden, line 17: Can we also do this on Travis too, please? It could either be added to the buildout configuration, or [directly in .travis.yml](https://github.com/zopefoundation/zope.configuration/blob/ce25c151d08579690c0e75e9cb88909a8d3c6068/.travis.yml#L4-L6).

src/ZODB/blob.py
```diff
@@ -733,7 +733,6 @@ def temporaryDirectory(self):
         return self.fshelper.temp_dir
 
 
-@zope.interface.implementer(ZODB.interfaces.IBlobStorage)
```
jamadden, line 736: I didn't encounter this as being inconsistent when I was looking at https://github.com/zopefoundation/zope.interface/issues/207. Did you find issues with it? 

I tend to like the static definition, even though the class itself can't be validated as such. What if the `directlyProvides` block in `__init__` was only executed in the `IBlobStorageRestorable` case? (Doesn't that interface extend `IBlobStorage`?)

Not super important though.

src/ZODB/FileStorage/FileStorage.py
```diff
@@ -127,12 +127,12 @@ def __init__(self, afile):
         self._file = afile
 
 @implementer(
-        IStorage,
         IStorageRestoreable,
         IStorageIteration,
         IStorageUndoable,
         IStorageCurrentRecordIteration,
         IExternalGC,
+        IStorage,
```
jamadden, line 135: `IStorageRestoreable` is definitely an extension of `IStorage`, as is `IStorageUndoable`, so to me it seems redundant to leave `IStorage` in the list at all.

But then again, `IStorageIteration` is named similarly and *doesn't* extend `IStorage`, so people can disagree. I wonder what others think.

janjaapdriessen: I moved `IStorage` around until zope.interface IRO checking was happy, I am +0 with removing IStorage from the list.

tox.ini
janjaapdriessen: Done in 6dfe525403273139e8ed3fefb559d9b10fe9029a.

src/ZODB/blob.py
janjaapdriessen: > Did you find issues with it?

I dove into the ZODB package because [zope.session](https://github.com/zopefoundation/zope.session/pull/13) the strict iro checker complains about ZODB when I run the zope.session tests with strict iro. 

I like your suggestion of [not touching the static definition](14e73eb3d6c273fd3227885de9d83009ba3ebb02).

src/ZODB/FileStorage/FileStorage.py
jamadden: Sounds good to me!

janjaapdriessen:
> All Python 3 builds are failing because they are building the sphinx docs, and `j1m.autointerface` is using Sphinx internals that have changed. I would suggest moving to `repoze.sphinx.autointerface`. Maybe that can happen in a separate PR and then this be rebased once that's merged?

I did spelunking and found out that `j1m.autointerface` was a [temporary fork](https://bitbucket.org/j1m/sphinxautointerface/src/default/README.rst) of [sphinxcontrib.zopeext](https://pypi.org/project/sphinxcontrib-zopeext/). I updated the dependencies in 463fc5afaae2005df808e020dcdf6dcd1cb841f2 and am waiting to see whether Travis will like these changes.

PR #317, opened 2020-06-12
Add a change note for #280

jamadden:
`ConnectionPool` and `ConnectionPool.map` both had docstrings and were used by third-party code. #280 removed that method but didn't leave a note. People should be warned about this potentially breaking change or they might find 5.6.0 to be surprising.

Even if the better thing to do is to restore `map()` in 5.6.1, I think 5.6.0 still needs a note. It's an easy thing to fix, at least in the cases I've seen that broke so I'm not too worried about it, as long as there's a heads-up.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -52,6 +52,10 @@
 
 - Fix inconsistent resolution order with zope.interface v5.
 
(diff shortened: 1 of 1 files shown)


jamadden:
Thank you.

Issue #318, opened 2020-07-07
DemoStorage does not take whiteouts into account -> leading data corruption


navytux:
Hello up there. I've discovered that DemoStorage wrongly handles load for objects deleted in read-write overlay leading to data corruption. It can be fixed if we correct interface for `IStorage.loadBefore`. Please find details in example program provided below.

Kirill

/cc @jimfulton, @hannosch

---- 8< ---- ([bug-demostorage-delete](https://lab.nexedi.com/kirr/ZODB/blob/e384a252/bug-demostorage-delete.py))
```py
#!/usr/bin/env python
# -*- coding: utf-8 -*-
"""Program bug-demostorage-delete demonstrates that DemoStorage wrongly handles
load for objects deleted in read-write overlay:

If object is deleted, even though a whiteout is recorded in the overlay,
loadBefore can incorrectly return object data of previous object revision from
read-only base, leading to data corruption.

Example output::

    base:
    @03d9186bbe8b5188  obj<0000000000000001>  ->  int(0)
    @03d9186bbe8ce3ee  obj<0000000000000001>  ->  int(1)

    base+overlay:
    @03d9186bbe8feddd  obj<0000000000000001>  ->  int(2)
    @03d9186bbe9147bb  root['obj'] = ø

    GC:
    @03d9186bbe931b00  del obj<0000000000000001>

    remount base+overlay:
    @03d9186bbe8feddd  obj<0000000000000001>  ->  int(2)    # must be int(2)
    @03d9186bbe9147bb  obj<0000000000000001>  ->  int(2)    # must be int(2)  (garbage, but still in DB)
    @03d9186bbe931b00  obj<0000000000000001>  ->  int(1)    # <- WRONG: must be POSKeyError,  not int(1) from base

The bug could be fixed if we change IStorage.loadBefore(oid) to return

    (None, serial)      # serial = revision where object was deleted

instead of just

    None

for deleted records.

Please see https://lab.nexedi.com/kirr/ZODB/commit/a762e2f8 for details.
"""

from __future__ import print_function

from ZODB.FileStorage import FileStorage
from ZODB.DemoStorage import DemoStorage
from ZODB.Connection import TransactionMetaData
from ZODB import DB
from persistent import Persistent
import transaction
import os, os.path


class PInt(Persistent):
    def __init__(self, i):
        self.i = i
    def __str__(self):
        return "int(%d)" % self.i

def h(tid):
    return tid.encode('hex')

def dump(obj):
[...]

PR #319, opened 2020-07-16
interface: Require invalidations to be called with full set of objects and not to skip transactions

navytux:
Currently invalidate documentation is not clear whether it should be
called for every transaction and whether it should include full set of
objects created/modified by that transaction. Until now this was working
relatively well for the sole purpose of invalidating client ZEO cache,
because for that particular task it is relatively OK not to include just
created objects into invalidation messages, and even to completely skip
sending invalidation if transaction only create - not modify - objects.
Due to this fact the workings of the client cache was indifferent to the
ambiguity of the interface.

In 2016 skipping transactions with only created objects was reconsidered
as bug and fixed in ZEO5 because ZODB5 relies more heavily on MVCC
semantic and needs to be notified about every transaction committed to
storage to be able to properly update ZODB.Connection view:

https://github.com/zopefoundation/ZEO/commit/02943acd#diff-52fb76aaf08a1643cdb8fdaf69e37802L889-R834
https://github.com/zopefoundation/ZEO/commit/9613f09b

However just-created objects were not included into invalidation
messages until, hopefully, recently:

https://github.com/zopefoundation/ZEO/pull/160

As ZODB is started to be used more widely in areas where it was not
traditionally used before, the ambiguity in invalidate interface and the
lack of guarantees - for any storage - to be notified with full set of
information, creates at least the following problems:

- a ZODB client (not necessarily native ZODB/py client) can maintain
  raw cache for the storage. If such client tries to load an oid at
  database view when that object did not existed yet, gets "no object"
  reply and stores that information into raw cache, to properly invalidate
  the cache it needs an invalidation message from ZODB server that
  *includes* created object.

- tools like `zodb watch` [1,2,3] don't work properly (give incorrect output)
  if not all objects modified/created by a transaction are included into
  invalidation messages.

- similarly to `zodb watch`, a monitoring tool, that would want to be
  notified of all created/modified objects, won't see full
  database-change picture, and so won't work properly without knowing
  which objects were created.

- wendelin.core 2 - which builds data from ZODB BTrees and data objects
  into virtual filesystem - needs to get invalidation messages with both
  modified and created objects to properly implement its own lazy
  invalidation and isolation protocol for file blocks in OS cache: when
  a block of file is accessed, all clients, that have this block mmaped,
  need to be notified and asked to remmap that block into particular
  revision of the file depending on a client's view of the filesystem and
  database [4,5].

  To compute to where a client needs to remmap the block, WCFS server
  (that in turn acts as ZODB client wrt ZEO/NEO server), needs to be able
  to see whether client's view of the filesystem is before object creation
  (and then ask that client to pin that block to hole), or after creation
  (and then ask the client to pin that block to corresponding revision).

  This computation needs ZODB server to send invalidation messages in
  full: with both modified and just created objects.

Also:

- the property that all objects - both modified and just created -
  are included into invalidation messages is required and can help to
  remove `next_serial` from `loadBefore` return in the future.
  This, in turn, can help to do 2x less SQL queries in loadBefore for
  NEO and RelStorage (and maybe other storages too):
  https://github.com/zopefoundation/ZODB/issues/318#issuecomment-657685745

Current state of storages with respect to new requirements:

- ZEO: does not skip transactions, but includes only modified - not
  created - objects. This is fixed by https://github.com/zopefoundation/ZEO/pull/160

- NEO: already implements the requirements in full

- RelStorage: already implements the requirements in full, if I
  understand correctly:

  https://github.com/zodb/relstorage/blob/3.1.2-1-gaf57d6c/src/relstorage/adapters/poller.py#L28-L145

While editing invalidate documentation, use the occasion to document
recently added property that invalidate(tid) is always called before
storage starts to report its lastTransaction() ≥ tid - see 4a6b0283
(mvccadapter: check if the last TID changed without invalidation).

/cc @jimfulton, @jamadden, @jmuchemb, @vpelletier, @arnaud-fontaine, @gidzit, @klawlf82, @jwolf083

[1] https://lab.nexedi.com/kirr/neo/blob/049cb9a0/go/zodb/zodbtools/watch.go
[2] https://lab.nexedi.com/kirr/neo/commit/e0d59f5d
[3] https://lab.nexedi.com/kirr/neo/commit/c41c2907

[4] https://lab.nexedi.com/kirr/wendelin.core/blob/1efb5876/wcfs/wcfs.go#L94-182
[5] https://lab.nexedi.com/kirr/wendelin.core/blob/1efb5876/wcfs/client/wcfs.h#L20-71

### DIFF
--- a/src/ZODB/interfaces.py
+++ b/src/ZODB/interfaces.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 ##############################################################################
 #
 # Copyright (c) Zope Corporation and Contributors.
@@ -280,9 +281,9 @@ class IStorageWrapper(Interface):
     - Out-of-band invalidation support
 
       A storage can notify it's wrapper of object invalidations that
-      don't occur due to direct operations on the storage.  Currently
-      this is only used by ZEO client storages to pass invalidation
-      messages sent from a server.
+      don't occur due to direct operations on the storage. This is used
+      by the client part of non-IMVCCStorage storages like ZEO and NEO
+      to pass invalidation messages sent from a storage server.
 
     - Record-reference extraction
 
@@ -317,6 +318,15 @@ class IStorageWrapper(Interface):
         """Invalidate object ids committed by the given transaction
 
         The oids argument is an iterable of object identifiers.
+
+        Unless the cache needs to be completely cleared via
+        invalidateCache event, the wrapped storage calls invalidate for
+        all transactions in the order as they are committed. For every
+        transaction the full set of its objects - both modified and just
+        created - is reported.
+
+        invalidate(tid) is always called before the storage starts to
+        report its lastTransaction() ≥ tid.
 
         The version argument is provided for backward
         compatibility. If passed, it must be an empty string.
@@ -1209,13 +1219,18 @@ class IMVCCStorage(IStorage):
     def poll_invalidations():
         """Poll the storage for external changes.
 
-        Returns either a sequence of OIDs that have changed, or None.  When a
-        sequence is returned, the corresponding objects should be removed
-        from the ZODB in-memory cache.  When None is returned, the storage is
-        indicating that so much time has elapsed since the last poll that it
-        is no longer possible to enumerate all of the changed OIDs, since the
-        previous transaction seen by the connection has already been packed.
-        In that case, the ZODB in-memory cache should be cleared.
+        Returns either None, or a sequence of OIDs with the full set of
+        objects that have been created or changed since the previous
+        call to poll_invalidations.
+
+        When a sequence is returned, the corresponding objects should be
+        removed from the ZODB in-memory cache.
+
+        When None is returned, the storage is indicating that so much time has
+        elapsed since the last poll that it is no longer possible to enumerate
+        all of the changed OIDs, since the previous transaction seen by the
+        connection has already been packed. In that case, the ZODB in-memory
+        cache should be cleared.
         """
 
     def sync(force=True):


navytux:
> Also:
>
> - the property that all objects - both modified and just created -
>   are included into invalidation messages is required and can help to
>   remove `next_serial` from `loadBefore` return in the future.
>   This, in turn, can help to do 2x less SQL queries in loadBefore for
>   NEO and RelStorage (and maybe other storages too):
>   https://github.com/zopefoundation/ZODB/issues/318#issuecomment-657685745

Corresponding patch is here: https://github.com/zopefoundation/ZODB/pull/323.

navytux:
> Current state of storages with respect to new requirements:
>
> - ZEO: does not skip transactions, but includes only modified - not
>   created - objects. This is fixed by https://github.com/zopefoundation/ZEO/pull/160
>
> - NEO: already implements the requirements in full
>
> - RelStorage: already implements the requirements in full, if I
>   understand correctly:
>
>  https://github.com/zodb/relstorage/blob/3.1.2-1-gaf57d6c/src/relstorage/adapters/poller.py#L28-L145

ZEO has just been fixed by https://github.com/zopefoundation/ZEO/commit/ab86bd72 (https://github.com/zopefoundation/ZEO/pull/160). Now it should thus be safe to merge this patch to interface.

Is it ok?

/cc @jamadden, @jimfulton

src/ZODB/interfaces.py
```diff
@@ -280,9 +281,9 @@ class IStorageWrapper(Interface):
     - Out-of-band invalidation support
 
       A storage can notify it's wrapper of object invalidations that
-      don't occur due to direct operations on the storage.  Currently
-      this is only used by ZEO client storages to pass invalidation
-      messages sent from a server.
+      don't occur due to direct operations on the storage. This is
+      used by non-IMVCCStorage storages like ZEO and NEO to pass
```
dataflake, line 285: `storages` should be `client storages`

src/ZODB/interfaces.py
```diff
@@ -280,9 +281,9 @@ class IStorageWrapper(Interface):
...
+      invalidation messages sent from storage server.
```
dataflake, line 286: "sent from __a__ storage server"

src/ZODB/interfaces.py
```diff
@@ -318,6 +319,14 @@ def invalidate(transaction_id, oids):
 
         The oids argument is an iterable of object identifiers.
 
+        Unless cache needs to be completely cleared via invalidateCache
```
dataflake, line 322: "__the__ cache"

src/ZODB/interfaces.py
```diff
@@ -318,6 +319,14 @@ def invalidate(transaction_id, oids):
...
+        event, wrapped storage calls invalidate for all transactions in
```
dataflake, line 323: "__the__ wrapped storage"

src/ZODB/interfaces.py
```diff
@@ -318,6 +319,14 @@ def invalidate(transaction_id, oids):
...
+        the order as they are committed. For every transaction full set
```
dataflake, line 324: "the order they are committed. For every transaction __the__ full set"

src/ZODB/interfaces.py
```diff
@@ -318,6 +319,14 @@ def invalidate(transaction_id, oids):
...
+        of its objects - both modified and just created - is reported.
+
+        invalidate(tid) is always called before storage starts to report
```
dataflake, line 327: "__the__ storage"

src/ZODB/interfaces.py
```diff
@@ -1209,8 +1218,8 @@ def release():
     def poll_invalidations():
         """Poll the storage for external changes.
 
-        Returns either a sequence of OIDs that have changed, or None.  When a
-        sequence is returned, the corresponding objects should be removed
+        Returns either None, or a sequence of OIDs that have created and changed.
```
dataflake, line 1221: "OIDs that have __been__ created __or__ changed."

src/ZODB/interfaces.py
navytux: Thanks for pointing this out. I reworked it a bit differently as _"This is used by **client part** of non-IMVCCStorage storages like ZEO and NEO to pass invalidation messages sent from **a** storage server."_

The reason for this language is that in ZODB speak IStorage already represents a client, not storage server. I hope it is ok.

src/ZODB/interfaces.py
navytux: ok

src/ZODB/interfaces.py
navytux: ok

src/ZODB/interfaces.py
navytux: ok

src/ZODB/interfaces.py
navytux: ok

src/ZODB/interfaces.py
navytux: ok

src/ZODB/interfaces.py
navytux: ok, but added "the full set" and reference to "previous call to poll invalidations":

_"Returns either None, or a sequence of OIDs **with the full set of objects** that have been created or changed **since previous call to poll_invalidations**."_

navytux:
@dataflake, thanks for feedback. I'm not a native speaker and am out of spoken language practice for a long time. Thanks for pointing out the places to fix. I took your input into account and amended the patch correspondingly. Is it better now?

<details>
<summary>interdiff</summary>

```diff
diff --git a/src/ZODB/interfaces.py b/src/ZODB/interfaces.py
index c2568c3c6..2e0dfea1a 100644
--- a/src/ZODB/interfaces.py
+++ b/src/ZODB/interfaces.py
@@ -281,9 +281,9 @@ class IStorageWrapper(Interface):
     - Out-of-band invalidation support
 
       A storage can notify it's wrapper of object invalidations that
-      don't occur due to direct operations on the storage. This is
-      used by non-IMVCCStorage storages like ZEO and NEO to pass
-      invalidation messages sent from storage server.
+      don't occur due to direct operations on the storage. This is used
+      by client part of non-IMVCCStorage storages like ZEO and NEO to
+      pass invalidation messages sent from a storage server.
 
     - Record-reference extraction
 
@@ -319,13 +319,14 @@ def invalidate(transaction_id, oids):
 
         The oids argument is an iterable of object identifiers.
 
-        Unless cache needs to be completely cleared via invalidateCache
-        event, wrapped storage calls invalidate for all transactions in
-        the order as they are committed. For every transaction full set
-        of its objects - both modified and just created - is reported.
+        Unless the cache needs to be completely cleared via
+        invalidateCache event, the wrapped storage calls invalidate for
+        all transactions in the order as they are committed. For every
+        transaction the full set of its objects - both modified and just
+        created - is reported.
 
-        invalidate(tid) is always called before storage starts to report
-        its lastTransaction() ≥ tid.
+        invalidate(tid) is always called before the storage starts to
+        report its lastTransaction() ≥ tid.
 
         The version argument is provided for backward
         compatibility. If passed, it must be an empty string.
@@ -1218,13 +1219,18 @@ def release():
     def poll_invalidations():
         """Poll the storage for external changes.
 
-        Returns either None, or a sequence of OIDs that have created and changed.
-        When a sequence is returned, the corresponding objects should be removed
-        from the ZODB in-memory cache.  When None is returned, the storage is
-        indicating that so much time has elapsed since the last poll that it
-        is no longer possible to enumerate all of the changed OIDs, since the
-        previous transaction seen by the connection has already been packed.
-        In that case, the ZODB in-memory cache should be cleared.
+        Returns either None, or a sequence of OIDs with the full set of
+        objects that have been created or changed since previous call to
+        poll_invalidations.
+
+        When a sequence is returned, the corresponding objects should be
+        removed from the ZODB in-memory cache.
+
+        When None is returned, the storage is indicating that so much time has
+        elapsed since the last poll that it is no longer possible to enumerate
+        all of the changed OIDs, since the previous transaction seen by the
+        connection has already been packed. In that case, the ZODB in-memory
+        cache should be cleared.
         """
 
     def sync(force=True):
```
</details>

navytux:
( CI failures are unrelated: it is only _"AppVeyor:  Could not resolve host: github.com"_ ([example](https://ci.appveyor.com/project/jimfulton/zodb/builds/34371101/job/hmn6iswmqd9owvrw)) )

src/ZODB/interfaces.py
```diff
@@ -1209,13 +1219,18 @@ def release():
...
 
-        Returns either a sequence of OIDs that have changed, or None.  When a
-        sequence is returned, the corresponding objects should be removed
-        from the ZODB in-memory cache.  When None is returned, the storage is
-        indicating that so much time has elapsed since the last poll that it
-        is no longer possible to enumerate all of the changed OIDs, since the
-        previous transaction seen by the connection has already been packed.
-        In that case, the ZODB in-memory cache should be cleared.
+        Returns either None, or a sequence of OIDs with the full set of
+        objects that have been created or changed since previous call to
```
dataflake, line 1223: "since __the__ previous call"

navytux:
@dataflake, thanks again. Oh, those articles... I've amended the patch according to your suggestions. Thanks for the approval.

<details>
<summary>interdiff</summary>

```diff
--- a/COMMIT_MSG.a
+++ b/COMMIT_MSG.b
@@ -92,6 +92,8 @@ Date:   Thu Jul 16 11:23:45 2020 +0300
     (mvccadapter: check if the last TID changed without invalidation).
     
     /cc @jimfulton, @jamadden, @jmuchemb, @vpelletier, @arnaud-fontaine, @gidzit, @klawlf82, @jwolf083
+    /reviewed-on https://github.com/zopefoundation/ZODB/pull/319
+    /reviewed-by @dataflake
     
     [1] https://lab.nexedi.com/kirr/neo/blob/049cb9a0/go/zodb/zodbtools/watch.go
     [2] https://lab.nexedi.com/kirr/neo/commit/e0d59f5d

--- a/src/ZODB/interfaces.py
+++ b/src/ZODB/interfaces.py
@@ -282,8 +282,8 @@ class IStorageWrapper(Interface):
 
       A storage can notify it's wrapper of object invalidations that
       don't occur due to direct operations on the storage. This is used
-      by client part of non-IMVCCStorage storages like ZEO and NEO to
-      pass invalidation messages sent from a storage server.
+      by the client part of non-IMVCCStorage storages like ZEO and NEO
+      to pass invalidation messages sent from a storage server.
 
     - Record-reference extraction
 
@@ -1220,8 +1220,8 @@ def poll_invalidations():
         """Poll the storage for external changes.
 
         Returns either None, or a sequence of OIDs with the full set of
-        objects that have been created or changed since previous call to
-        poll_invalidations.
+        objects that have been created or changed since the previous
+        call to poll_invalidations.
 
         When a sequence is returned, the corresponding objects should be
         removed from the ZODB in-memory cache.
```
</details>

navytux:
( CI is ok again, besides _"AppVeyor: Failed to connect to github.com port 443:"_ once )

dataflake:
There's no need to explain every single CI failure, especially if it's obvious the tool is at fault. The reviewer's job is to look at those failures and recognize tool failures - or spurious test failures because the tests themselves are not reliable.

navytux:
@dataflake, thanks. I'm used to situation where reviewers do not even take a look unless tests pass, that's why I point out that CI failures are unrelated all the time. In my understanding it is noone's "job" to look into my changes, that's why I try to present it in clean and easy-to-review state. It would be, of course, good to generally improve continuous integration and remove flakiness.

_Using the occassion: I would indeed be grateful if someone can take a look at technical bits. Or is it all ok, given that ZEO is now fixed and so all the storages, that are in use today, confirm to the amended interface definition?_

Kirill

dataflake:
> _Using the occassion: I would indeed be grateful if someone can take a look at technical bits. Or is it all ok, given that ZEO is now fixed and so all the storages, that are in use today, confirm to the amended interface definition?_

My experience has been that unless you specifically assign reviewers no one will feel responsible and no review happens. It's not enough to put CC `@` style mentions into the text. Assign people and, if necessary, prod them by adding comments.

navytux:
ok, thanks, assigning @jimfulton and @jamadden.

src/ZODB/interfaces.py
```diff
@@ -280,9 +281,9 @@ class IStorageWrapper(Interface):
     - Out-of-band invalidation support
 
       A storage can notify it's wrapper of object invalidations that
```
jmuchemb, line 283: s/it's/its/

jmuchemb (approved):

src/ZODB/interfaces.py
navytux: Yes, but I'm not changing this line at all and the "it's" come from old text. There are other usage of "it's" in `ZODB/interfaces.py` where I would put "its" by myself. I thought it is kind of US style or something similar and not delved into changing what is semantically not covered by the patch subject. If we are to change "it's" -> "its" I think it is better to do in a separate patch and for the whole file or whole codebase.

Thanks for the approval.

navytux:
If there is no objection I will merge on Monday Aug 31.

navytux:
Patch applied to master as c1e08052.

Issue #324, opened 2020-08-18
scripts/fsoids.py throws UnboundLocalError


krissik:
I'm hunting an error in my zodb. Because of that I need to know where an object is referenced. When I run scripts/fsoids.py I get the error

```
root@myhost:/opt/plone/worker0-zodbconvert-zodbverify-2020-05-18/instance# bin/zopepy /opt/plone/buildout-cache/eggs/ZODB-5.5.0-py2.7.egg/ZODB/scripts/fsoids.py /opt/plone/blobs/converted-to-filestorage/Data.fs 0x03f79c
Traceback (most recent call last):
  File "bin/zopepy", line 377, in <module>
    exec(compile(__file__f.read(), __file__, "exec"))
  File "/opt/plone/buildout-cache/eggs/ZODB-5.5.0-py2.7.egg/ZODB/scripts/fsoids.py", line 79, in <module>
    main()
  File "/opt/plone/buildout-cache/eggs/ZODB-5.5.0-py2.7.egg/ZODB/scripts/fsoids.py", line 75, in main
    c.run()
  File "/opt/plone/buildout-cache/eggs/ZODB-5.3.0-py2.7.egg/ZODB/FileStorage/fsoids.py", line 130, in run
    self._check_trec(txn)
  File "/opt/plone/buildout-cache/eggs/ZODB-5.3.0-py2.7.egg/ZODB/FileStorage/fsoids.py", line 143, in _check_trec
    self._check_drec(drec)
  File "/opt/plone/buildout-cache/eggs/ZODB-5.3.0-py2.7.egg/ZODB/FileStorage/fsoids.py", line 173, in _check_drec
    for ref, klass in get_refs(pick):
  File "/opt/plone/buildout-cache/eggs/ZODB-5.3.0-py2.7.egg/ZODB/serialize.py", line 696, in get_refs
    if not isinstance(oid, bytes):
UnboundLocalError: local variable 'oid' referenced before assignment
```

krissik:
I think #295 will fix it.

sallner:
Yes this is a duplicate of #285 and should be fixed by #295. Closing this issue.

PR #325, opened 2020-08-18
Correct variable name

krissik:
Fix typo that might be the cause of #324


krissik:
This duplicates #295 which I saw too late.

PR #329, opened 2020-09-22
fix variable name in get_refs function

dest81:
I am getting error: local variable 'oid' referenced before assignment if reference is not tuple

```
for reference in refs:
        if isinstance(reference, tuple):
            oid, klass = reference
        elif isinstance(reference, (bytes, str)):
            data, klass = reference, None
        else:
            assert isinstance(reference, list)
            continue

        if not isinstance(oid, bytes):
            assert isinstance(oid, str)
            # this happens on Python 3 when all bytes in the oid are < 0x80
            oid = oid.encode('ascii')
```

### DIFF
--- a/src/ZODB/serialize.py
+++ b/src/ZODB/serialize.py
@@ -699,7 +699,7 @@ def get_refs(a_pickle):
         if isinstance(reference, tuple):
             oid, klass = reference
         elif isinstance(reference, (bytes, str)):
-            data, klass = reference, None
+            oid, klass = reference, None
         else:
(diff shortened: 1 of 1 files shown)


mgedmin:
Duplicate of #295, which was never merged for silly reasons (CI was broken back then).

Could use a changelog note.

mgedmin:
I merged #295 since it had a changelog entry and got rebased recently to pass CI.  I'm sure that wouldn't have happened if you hadn't opened this PR, so thank you very much for that!

PR #333, opened 2021-01-22
Update tutorial.rst

akilmarshall:
sample code was incorrect.

### DIFF
--- a/doc/tutorial.rst
+++ b/doc/tutorial.rst
@@ -83,7 +83,7 @@ storage, a database and finally a connection::
     storage = ZODB.FileStorage.FileStorage('mydata.fs')
     db = ZODB.DB(storage)
     connection = db.open()
-    root = connection.root
+    root = connection.root()
 
 ZODB has a pluggable storage framework.  This means there are a
 variety of storage implementations to meet different needs, from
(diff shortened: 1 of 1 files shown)


doc/tutorial.rst
```diff
@@ -83,7 +83,7 @@ storage, a database and finally a connection::
     storage = ZODB.FileStorage.FileStorage('mydata.fs')
     db = ZODB.DB(storage)
     connection = db.open()
-    root = connection.root
+    root = connection.root()
```
jamadden, line 86:
```suggestion
    root = connection.root
```

I believe the code was correct as it was. The intent here is to store objects in the root dictionary as key/value pairs, not to add new *attributes* to the root dictionary object.

`connection.root` returns a "convenience" object that turns attribute access into dictionary key storage. Thus, `connection.root.foo = 1` is the same thing as `connection.root()['foo'] = 1`. But `connection.root()` returns the actual dict, and `connection.root().foo = 1` is *not* the same thing as `connection.root()['foo'] = 1`.

```pycon
>>> type(connection.root)
ZODB.Connection.RootConvenience
>>> type(connection.root())
persistent.mapping.PersistentMapping
>>> connection.root.foo = 1
>>> connection.root()
 {'foo': 1}
>>> connection.root().biz = 42
>>> connection.root()
 {'foo': 1}
```

jamadden (changes requested):

doc/tutorial.rst
akilmarshall: Ok, sorry for wasting your time.

PR #338, opened 2021-03-19
WIP: fsrefs: Optimize IO

navytux:
Update 2021-03-25: This patch turned out to require too much RAM (see https://github.com/zopefoundation/ZODB/pull/338#issuecomment-804539929 and below).
https://github.com/zopefoundation/ZODB/pull/340 is corrected version that still optimizes fsrefs, but does not require much RAM to work.

--------

- access objects in the order of their position in file
- load objects only once

This should give dramatical speedup when data are on HDD.

For example @perrinjerome reports that on a 73Go database it takes
almost 8h to run fsrefs (where on the same database, fstest takes 15
minutes). After the patch fsrefs took ~74 minutes to run on the same
database. In other words this is ~ 6.5x improvement.

Fsrefs has no tests. I tested it only lightly via generating a bit
corrupt database with deleted referred object(*), and it gives the same
output as unmodified fsrefs.

    oid 0x0 __main__.Object
    last updated: 1979-01-03 21:00:42.900001, tid=0x285cbacb70a3db3
    refers to invalid objects:
            oid 0x07 missing: '<unknown>'
            oid 0x07 object creation was undone: '<unknown>'

[1] https://lab.nexedi.com/nexedi/zodbtools/merge_requests/19#note_129480
[2] https://lab.nexedi.com/nexedi/zodbtools/merge_requests/19#note_129551

(*) test database generated via a bit modified gen_testdata.py from
zodbtools:

https://lab.nexedi.com/nexedi/zodbtools/blob/v0.0.0.dev8-28-g129afa6/zodbtools/test/gen_testdata.py +

```diff
--- a/zodbtools/test/gen_testdata.py
+++ b/zodbtools/test/gen_testdata.py
@@ -229,7 +229,7 @@ def ext(subj): return {}
         # delete an object
         name = random.choice(list(root.keys()))
         obj = root[name]
-        root[name] = Object("%s%i*" % (name, i))
+#       root[name] = Object("%s%i*" % (name, i))
         # NOTE user/ext are kept empty on purpose - to also test this case
         commit(u"", u"predelete %s" % unpack64(obj._p_oid), {})
```

/cc @tim-one, @jeremyhylton

### DIFF
--- a/src/ZODB/scripts/fsrefs.py
+++ b/src/ZODB/scripts/fsrefs.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
 ##############################################################################
 #
 # Copyright (c) 2002 Zope Foundation and Contributors.
@@ -62,14 +61,19 @@ of objects, it does not attempt to load objects in versions, or non-current
 revisions of objects; therefore fsrefs cannot find problems in versions or
 in non-current revisions.
 """
+
+# implementation note: to save RAM we work with OIDs as with 64-bit integers
+# and keep references graph in LOBTree.
+
 from __future__ import print_function
 import traceback
 
 from ZODB.FileStorage import FileStorage
 from ZODB.TimeStamp import TimeStamp
-from ZODB.utils import u64, oid_repr, get_pickle_metadata, load_current
+from ZODB.utils import u64, p64, oid_repr, get_pickle_metadata, load_current
 from ZODB.serialize import get_refs
 from ZODB.POSException import POSKeyError
+from BTrees.LOBTree import LOBTree, TreeSet as LTreeSet
 
 # There's a problem with oid.  'data' is its pickle, and 'serial' its
 # serial number.  'missing' is a list of (oid, class, reason) triples,
@@ -81,7 +85,7 @@ def report(oid, data, serial, missing):
     else:
         plural = ""
     ts = TimeStamp(serial)
-    print("oid %s %s.%s" % (hex(u64(oid)), from_mod, from_class))
+    print("oid %s %s.%s" % (hex(oid), from_mod, from_class))
     print("last updated: %s, tid=%s" % (ts, hex(u64(serial))))
     print("refers to invalid object%s:" % plural)
     for oid, info, reason in missing:
@@ -89,7 +93,7 @@ def report(oid, data, serial, missing):
             description = "%s.%s" % info
         else:
             description = str(info)
-        print("\toid %s %s: %r" % (oid_repr(oid), reason, description))
+        print("\toid %s %s: %r" % (oid_repr(p64(oid)), reason, description))
     print()
 
 def main(path=None):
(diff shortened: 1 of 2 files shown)


src/ZODB/scripts/fsrefs.py
```diff
@@ -118,30 +119,38 @@ def main(path=None):
     # This does not include oids in undone.
     noload = {}
 
-    for oid in fs._index.keys():
+    #print("# building references graph ...")
+    graph = {} # oid -> refs
+
+    posoidv = list((pos, oid) for (oid, pos) in fs._index.items()) # [] of (pos, oid)
+    posoidv.sort() # access objects in order of pos↑  (optimize disk IO)
```
jamadden, line 126: Can we avoid the use of unnecessary non-ascii characters?
```suggestion
    posoidv.sort() # access objects in order of ascending file position  (optimize disk IO)
```

src/ZODB/scripts/fsrefs.py
```diff
@@ -118,30 +119,38 @@ def main(path=None):
...
-        if oid in inactive:
             continue
-        data, serial = load_current(fs, oid)
+
         refs = get_refs(data)
+        graph[oid] = refs
+
+    #print("# verifying reachability ...")
+    oidv = list(graph.keys())
+    oidv.sort() # verify objects in order of oid↑  (useful for human perception; stable output)
```
jamadden, line 146:
```suggestion
    oidv.sort() # verify objects in order of ascending oid  (useful for human perception; stable output)
```

jamadden (commented):
I don't use fsrefs, so I can't comment on the functionality, but I like the concept of ordering the IO, and I don't notice anything obviously wrong.

I am slightly concerned about the increased memory usage; that appears to be the major tradeoff of keeping everything in `graph` instead of loading again. For very large databases that might be a problem? How much memory does this take on the example 73GB database? That could possibly be optimized by using, say, an `array.array` of 64-bit integers instead of a list as the values of the `graph`, and possibly a bit more by using a `LOTree` for graph itself.

I think this deserves a change note.

src/ZODB/scripts/fsrefs.py
navytux: ok

src/ZODB/scripts/fsrefs.py
navytux: ok

navytux:
@jamadden, thanks for feedback. I've changed fsrefs back not to use non-ascii. Using `LOBTree` also makes sense, as you say, to save RAM. This way I've changed fsrefs to universally work with object IDs as with 64-bit integer which brought other RAM-savings like changing `noload` and `undone` from dict to `LTreeSet`. Currently I don't have access to test database in question, so maybe @perrinjerome might recheck how much RAM is consumed by the graph. I don't expect it to be too much however, even if graph is potentially O(n^2), in practice most Persistent objects refer to limited number of other objects, so chances are the graph RAM consumption is closer to N(n^{1+ε}). Anyway using `LOBTree` should be universally good (and it automatically provides us with ordered oid traversal, instead of another list and sorting), and we can apply other RAM-saving optimizations later, if RAM consumption would turn to indeed be very high.

navytux:
( I pushed updates as follow-up patches according to your review preferrence )

jamadden:
> This way I've changed fsrefs to universally work with object IDs as with 64-bit integer which brought other RAM-savings like changing noload and undone from dict to LTreeSet

Nice!

navytux:
> I completely forgot to mention: OIDs are unsigned 64-bit quantities, right?

I thought about it too, and you are right - even though [max tid is 2^63-1](https://github.com/zopefoundation/ZODB/blob/5.6.0-15-g22d1405d4/src/ZODB/utils.py#L80), there is nothing saying that about oid. So let's indeed switch to QOBtree flavour.

navytux:
> Nice!

Thanks

perrinjerome:
Thank you @navytux and @jamadden 

To measure memory usage, I ran a command like this which uses ps to capture rss (and vsz) every 10 seconds:

```bash
#!bin/bash

WATCHED_PID=$({ python src/ZODB/scripts/fsrefs.py $1 >log.stdout 2>log.stderr & } && echo $!);
PS_COLUMNS="etime %cpu %mem rss vsz"
echo "      "$PS_COLUMNS
while ps -p $WATCHED_PID --no-headers --format "$PS_COLUMNS"; do 
   sleep 10
done
```

The maximum rss and the execution time ( in hour:minute:seconds) was this:

| commit | maximum (rss) memory usage | total execution time |  
| --- | --- | --- | 
| 22d1405d (before) | 5509040.0 ( 5.25Go ) | 07:30:25 | 
| 2f638adf1 (first version of the patch) | 38651384.0 ( 36.86Go ) | 01:11:47 |
| fe3ecda3 (current version of the patch, with optimized memory usage) | 29948784.0 ( 28.56Go ) | 01:16:27 |

For reference, the measurement files:

[measure_22d1405d.txt](https://github.com/zopefoundation/ZODB/files/6186520/measure_22d1405d.txt)
[measure_2f638adf1.txt](https://github.com/zopefoundation/ZODB/files/6186517/measure_2f638adf1.txt)
[measure_fe3ecda3.txt](https://github.com/zopefoundation/ZODB/files/6186519/measure_fe3ecda3.txt)

navytux:
Thanks, @jerome. The memory consumption is too much. I will provide more updates today.

navytux:
I've tried to optmize memory consumption more:

- use QQBTree instead of [](pos,oid) for pos -> oid index.
  The list was using ~ 10GB of RAM, QQBTree uses ~ 2GB.
- don't remember ref's klass in the graph
- use QSet<oid> for refs in the graph

but still RAM usage is too much - ~15GB. Please see details in https://github.com/zopefoundation/ZODB/pull/338/commits/8a09427b.

I think it makes sense to retry a second version of this patch that preserves 2-pass aproach instead of building in-memory graph, and only changes iteration from oid-ordered to pos-ordered.

navytux:
> I think it makes sense to retry a second version of this patch that preserves 2-pass aproach instead of building in-memory graph, and only changes iteration from oid-ordered to pos-ordered.

Please see this follow-up here: https://github.com/zopefoundation/ZODB/pull/340.

navytux:
https://github.com/zopefoundation/ZODB/pull/340 was landed instead.

PR #341, opened 2021-03-26
DemoStorage: Add support for deleteObject (IExternalGC)

navytux:
So that it is generally possible to zodbrestore[1,2] zodbdumped[3]
database/transactions into DemoStorage. Because without this patch, when dump
contains deletion records, e.g.

    txn 0285cbacc06d3a4c " " 
    user ""
    description ""
    extension ""
    obj 0000000000000007 delete

it fails like this:

      ...
      File "/home/kirr/src/wendelin/z/zodbtools/zodbtools/zodbrestore.py", line 94, in main
        zodbrestore(stor, asbinstream(sys.stdin), _)
      File "/home/kirr/src/wendelin/z/zodbtools/zodbtools/zodbrestore.py", line 43, in zodbrestore
        zodbcommit(stor, at, txn)
      File "/home/kirr/src/wendelin/z/zodbtools/zodbtools/zodbcommit.py", line 131, in zodbcommit
        _()
      File "/home/kirr/src/wendelin/z/zodbtools/zodbtools/zodbcommit.py", line 118, in _
        stor.deleteObject(obj.oid, current_serial(obj.oid), txn)
    AttributeError: 'DemoStorage' object has no attribute 'deleteObject'
        demo_test.go:105: demo:(file:///tmp/demo470143633/δ0285cbac258bf266/base.fs)/(file:///tmp/demo470143633/δ0285cbac258bf266/δ.fs): zpyrestore: exit status 1

To be able to implement DemoStorage.deleteObject, we have to fix 
FileStorage.deleteObject first: currently FileStorage.deleteObject raises
POSKeyError if an object is not present in the database, but for situation
where

- demo=base+δ,
- object is present in base, and 
- object is not present in δ

it creates a problem because there is no way to add whiteout record for the 
object into δ.

This change should be generally good; let me quote
https://lab.nexedi.com/kirr/neo/commit/543041a3 for why:

---- 8< ----
Even though currently it is not possible to create such data record via 
FileStorage(py).deleteObject (because it raises POSKeyError if there is
no previous object revision), being able to use such data records is
semantically useful in overlayed DemoStorage settings, where δ part
marks an object that exists only in base with delete record whiteout.

It is also generally meaningful to be able to create "delete" record
even if object was not previously existing: "deleteObject" is actually
similar to "store" (and so should be better named as "storeDelete"). If
one wants to store deletion, there should not be a reason to reject it, 
because deleteObject already has seatbelt in the form of oldserial, and 
if the user calls deleteObject(oid, oldserial=z64), he/she is already
telling that "I know that this object does not exist in this storage
(oldserial=z64), but still please create a deletion record for it". Once
again this is useful in overlayed DemoStorage settings described above.

For the reference, such whiteout deletion records pass ZODB/scripts/fstest just fine.
---- 8< ----

DemoStorage.deleteObject implementation is straightforward, but builds on
loadAt[4] as precondition.

/cc @jimfulton, @jamadden, @perrinjerome

[1] https://lab.nexedi.com/nexedi/zodbtools/blob/129afa67/zodbtools/zodbrestore.py
[2] https://lab.nexedi.com/nexedi/zodbtools/merge_requests/19
[3] https://lab.nexedi.com/nexedi/zodbtools/blob/129afa67/zodbtools/zodbdump.py
[4] https://github.com/zopefoundation/ZODB/pull/323

### DIFF
--- a/src/ZODB/BaseStorage.py
+++ b/src/ZODB/BaseStorage.py
@@ -59,7 +59,7 @@ class BaseStorage(UndoLogCompatible):
 
     If it stores multiple revisions, it should implement
     loadSerial()
-    loadBefore()
+    loadAt()
 
     Each storage will have two locks that are accessed via lock
     acquire and release methods bound to the instance.  (Yuck.)
@@ -267,9 +267,8 @@ class BaseStorage(UndoLogCompatible):
         raise POSException.Unsupported(
             "Retrieval of historical revisions is not supported")
 
-    def loadBefore(self, oid, tid):
-        """Return most recent revision of oid before tid committed."""
-        return None
+    # do not provide loadAt/loadBefore here in BaseStorage - if child forgets
+    # to override it - storage will always return "no data" instead of failing.
 
     def copyTransactionsFrom(self, other, verbose=0):
         """Copy transactions from another storage.
--- a/src/ZODB/DB.py
+++ b/src/ZODB/DB.py
@@ -726,8 +726,7 @@ class DB(object):
           - `before`: like `at`, but opens the readonly state before the
             tid or datetime.
         """
-        # `at` is normalized to `before`, since we use storage.loadBefore
-        # as the underlying implementation of both.
+        # `at` is normalized to `before`.
         before = getTID(at, before)
         if (before is not None and
             before > self.lastTransaction() and
--- a/src/ZODB/DemoStorage.py
+++ b/src/ZODB/DemoStorage.py
@@ -24,6 +24,7 @@ import os
 import random
 import weakref
 import tempfile
+import warnings
 import ZODB.BaseStorage
 import ZODB.blob
 import ZODB.interfaces
@@ -217,45 +218,55 @@ class DemoStorage(ConflictResolvingStorage):
     # still want load for old clients (e.g. zeo servers)
     load = load_current
 
-    def loadBefore(self, oid, tid):
-        try:
-            result = self.changes.loadBefore(oid, tid)
-        except ZODB.POSException.POSKeyError:
-            # The oid isn't in the changes, so defer to base
-            return self.base.loadBefore(oid, tid)
+    def loadAt(self, oid, at):
+        data, serial = ZODB.utils.loadAt(self.changes, oid, at)
+        if (data is not None) or (serial != ZODB.utils.z64):
+            # object is present in changes either as data or deletion record.
+            return data, serial
 
-        if result is None:
-            # The oid *was* in the changes, but there aren't any
-            # earlier records. Maybe there are in the base.
-            try:
-                result = self.base.loadBefore(oid, tid)
-            except ZODB.POSException.POSKeyError:
-                # The oid isn't in the base, so None will be the right result
-                pass
+        # object is not present in changes at all - use base
+        return ZODB.utils.loadAt(self.base, oid, at)
+
+    def loadBefore(self, oid, before):
+        warnings.warn("loadBefore is deprecated - use loadAt instead",
+                DeprecationWarning, stacklevel=2)
+        p64 = ZODB.utils.p64
+        u64 = ZODB.utils.u64
+
+        if before in (maxtid, ZODB.utils.z64):
+            at = before
+        else:
+            at = p64(u64(before)-1)
+
+        data, serial = self.loadAt(oid, at)
+
+        # find out next_serial.
+        # it is ok to use dumb/slow implementation since loadBefore should not
+        # be used and is provided only for backward compatibility.
+        next_serial = maxtid
+        while 1:
+            _, s = self.loadAt(oid, p64(u64(next_serial)-1))
+            assert s >= serial
+            if s == serial:
+                # found - next_serial is serial of the next data record
+                break
+            next_serial = s
+
+        if next_serial == maxtid:
+            next_serial = None
+
+        # next_serial found -> return/raise what loadBefore users expect
+        if data is None:
+            if next_serial is None:
+                # object was never created
+                raise ZODB.POSException.POSKeyError(oid)
             else:
-                if result and not result[-1]:
-                    # The oid is current in the base.  We need to find
-                    # the end tid in the base by fining the first tid
-                    # in the changes. Unfortunately, there isn't an
-                    # api for this, so we have to walk back using
-                    # loadBefore.
+                # object was deleted
+                return None
 
-                    if tid == maxtid:
-                        # Special case: we were looking for the
-                        # current value. We won't find anything in
-                        # changes, so we're done.
-                        return result
+        # regular data record
+        return data, serial, next_serial
 
-                    end_tid = maxtid
-                    t = self.changes.loadBefore(oid, end_tid)
-                    while t:
-                        end_tid = t[1]
-                        t = self.changes.loadBefore(oid, end_tid)
-                    result = result[:2] + (
-                        end_tid if end_tid != maxtid else None,
-                        )
-
-        return result
 
     def loadBlob(self, oid, serial):
         try:
--- a/src/ZODB/DemoStorage.py
+++ b/src/ZODB/DemoStorage.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 ##############################################################################
 #
 # Copyright (c) Zope Corporation and Contributors.
@@ -108,10 +109,15 @@ class DemoStorage(ConflictResolvingStorage):
             if close_changes_on_close is None:
                 close_changes_on_close = False
         else:
+            self._temporary_changes = False
             if ZODB.interfaces.IBlobStorage.providedBy(changes):
                 zope.interface.alsoProvides(self, ZODB.interfaces.IBlobStorage)
             if close_changes_on_close is None:
                 close_changes_on_close = True
+
+        if ZODB.interfaces.IExternalGC.providedBy(changes):
+            zope.interface.alsoProvides(self, ZODB.interfaces.IExternalGC)
+            self.deleteObject = self._storeDelete
 
         self.changes = changes
         self.close_changes_on_close = close_changes_on_close
@@ -375,6 +381,40 @@ class DemoStorage(ConflictResolvingStorage):
             self._resolved.append(oid)
         else:
             self.changes.store(oid, serial, data, '', transaction)
+
+    # _storeDelete serves deleteObject when .changes implements IExternalGC.
+    def _storeDelete(self, oid, oldserial, transaction):
+        if transaction is not self._transaction:
+            raise ZODB.POSException.StorageTransactionError(self, transaction)
+
+        # oldserial ∈ changes  -> changes.deleteObject
+        baseHead = self.base.lastTransaction()
+        if oldserial > baseHead:
+            self.changes.deleteObject(oid, oldserial, transaction)
+            return
+
+        # oldserial ∈ base  -> find out it is indeed the latest there and then
+        # call changes.deleteObject(oldserial=z64)
+
+        _, serial = ZODB.utils.loadAt(self.changes, oid, self.changes.lastTransaction())
+        if serial != ZODB.utils.z64:
+            # object has data or deletion record in changes
+            raise ZODB.POSException.ConflictError(oid=oid, serials=(serial, oldserial))
+
+        _, serial = ZODB.utils.loadAt(self.base, oid, baseHead)
+        if serial != oldserial:
+            raise ZODB.POSException.ConflictError(oid=oid, serials=(serial, oldserial))
+
+        # object has no data/deletion record in changes and its latest revision in base == oldserial
+        # -> changes.deleteObject(oldserial=z64) + correct oldserial back on conflict.
+        try:
+            self.changes.deleteObject(oid, ZODB.utils.z64, transaction)
+        except ZODB.POSException.ConflictError as e:
+            assert len(e.serials) == 2
+            assert e.serials[1] == ZODB.utils.z64
+            e.serials = (e.serials[0], oldserial)
+            raise
+
 
     def storeBlob(self, oid, oldserial, data, blobfilename, version,
                   transaction):
--- a/src/ZODB/FileStorage/FileStorage.py
+++ b/src/ZODB/FileStorage/FileStorage.py
@@ -661,9 +661,10 @@ class FileStorage(
         with self._lock:
             old = self._index_get(oid, 0)
             if not old:
-                raise POSKeyError(oid)
-            h = self._read_data_header(old, oid)
-            committed_tid = h.tid
+                committed_tid = z64
+            else:
+                h = self._read_data_header(old, oid)
+                committed_tid = h.tid
 
             if oldserial != committed_tid:
                 raise ConflictError(
--- a/src/ZODB/FileStorage/FileStorage.py
+++ b/src/ZODB/FileStorage/FileStorage.py
@@ -21,6 +21,7 @@ import errno
 import logging
 import os
 import time
+import warnings
 from struct import pack
 from struct import unpack
 
@@ -57,6 +58,7 @@ from ZODB.interfaces import IStorageCurrentRecordIteration
 from ZODB.interfaces import IStorageIteration
 from ZODB.interfaces import IStorageRestoreable
 from ZODB.interfaces import IStorageUndoable
+from ZODB.interfaces import IStorageLoadAt
 from ZODB.POSException import ConflictError
 from ZODB.POSException import MultipleUndoErrors
 from ZODB.POSException import POSKeyError
@@ -133,6 +135,7 @@ class TempFormatter(FileStorageFormatter):
         IStorageCurrentRecordIteration,
         IExternalGC,
         IStorage,
+        IStorageLoadAt,
         )
 class FileStorage(
     FileStorageFormatter,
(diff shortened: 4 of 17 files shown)


navytux:
Needs loadAt (https://github.com/zopefoundation/ZODB/pull/323) and so includes it here.
The change itself is in commit 49f5959e.
Probably WIP because testing coverage is not complete.

navytux:
See also https://lab.nexedi.com/kirr/neo/commit/4fb6bd0a?expanded=1 for the reference.

navytux:
I was running ZODB/go tests with forgetting to switch my ZODB/py checkout from master to my custom branch, and once again hit the problem of `AttributeError: 'DemoStorage' object has no attribute 'deleteObject'` as explained in the [description](https://github.com/zopefoundation/ZODB/pull/341#issue-601299591). So I recalled about this pull request.

So is there anyone with feedback maybe?
@jamadden, @d-maurer?

Once again, hereby patch builds on top of https://github.com/zopefoundation/ZODB/pull/323 which, by the way, fixes data corruption problem of DemoStorage (https://github.com/zopefoundation/ZODB/issues/318).

navytux:
Refreshed the code by syncing with https://github.com/zopefoundation/ZODB/pull/323 to resolve conflicts. As explained in the [main message](https://github.com/zopefoundation/ZODB/pull/341#issue-841583557), `DemoStorage.deleteObject` implementation is straightforward, but builds on loadAt (now loadBeforeEx) as precondition.

@d-maurer, please note how loadAt -> loadBeforeEx renaming made the code somewhat less clear: https://github.com/zopefoundation/ZODB/pull/341/commits/c3820ff48eda59f6efc14bedc1d9cc37b6d59c7d.

There, we are obtaining `base.lastTransaction()` and `changes.lastTransaction()` and then want to load data from storages _at_ that particular states. After at -> before transition, we cannot use that lastTransaction return as is and need to do `p64(u64(·)+1)` all the time around it.

To me it is a common situation - when something determines revision for at which database state it wants to observe data snapshot, and then accesses/loads data _at_ that particular revision. Frequently determining revision is done via `.lastTransaction()`, but it can be also done by other means. With loadAt accessing that snapshot is straightforward, while with loadBefore* we/users are forced to do `p64(u64(·)+1)` dance all the time.

I don't insist on this strongly, but maybe it would make sense to reconsider loadAt -> loadBeforeEx rename and rename `loadBeforeEx` back to -> `loadAt` before it is merged?

This will unify loading semantic with `loadSerial` (which loads _at_ particular revision).
And actually we can also deprecate and eventually remove `loadSerial`, because in practice, usage of `loadSerial` can be almost directly replaced with `loadAt`.

Kirill

navytux:
And I would appreciate if someone would look at the main patch in the PR: https://github.com/zopefoundation/ZODB/commit/49f5959e73e6ba0d78187d572275eade27b61fc8.

icemac:
@d-maurer Do you think you are able to review this PR? Or is there anyone else with deep enough ZODB knowledge?

d-maurer:
Michael Howitz wrote at 2022-3-17 00:15 -0700:
***@***.*** Do you think you are able to review this PR? Or is there anyone else with deep enough ZODB knowledge?

This PR depends on the `loadAt` PR (--> #323).
I have approved this PR (after long discussions).
But, @jamadden might have had some concerns
(--> "https://github.com/zopefoundation/ZODB/pull/323#issuecomment-955089963"),
in particular he was unsure about the impact on `RelStorage`.

PR #323 introduces a new storage method (`loadAt` or `loadBeforeEx`);
@navitux has implemented it for the storages directly provided
by `ZODB`; however other storages lack the new method and the
provided fallback might not be efficient.


If #323 is accepted, this PR is trivial and poses no further risk.

navytux:
@icemac, @d-maurer, thanks for heads up and for feedback.

This patch is indeed small, hopefully safe, change, but it requires https://github.com/zopefoundation/ZODB/pull/323 to work.

Regarding potential impact of https://github.com/zopefoundation/ZODB/pull/323 on RelStorage - I believe I've addressed that with https://github.com/zopefoundation/ZODB/pull/323#issuecomment-966641420 and https://github.com/zodb/relstorage/pull/485. However taking into account that @jamadden might be having a hard time, it is understandable that there is no feedback on this.

Another point is that personally, even though https://github.com/zopefoundation/ZODB/pull/323 is approved (@d-maurer thanks for the approval and for the discussion), I'm not so comfortable to merge it in loadBeforeEx form compared to merging it as loadAt. The reason here is that until we establish a public interface the freedom to change naming is still there, but once the interface is established it has to be kept in stone.

In my view loadAt is more descriptive and "speaking" name compared to loadBeforeEx. It is also more handy to use loadAt compared to loadBefore(Ex) because one does not need +1 transactionID obtained via e.g. lastTransaction. loadAt can also make loadSerial obsolete today, without any blockers.

Dieter, would you please try to review your decision to prefer loadBefore-style for new function compared to loadAt? I understand that you still might be keeping on your preferences, but maybe taking another look afresh might be helpful. Because you originally [asked](https://github.com/zopefoundation/ZODB/pull/323#pullrequestreview-650963363) to change loadAt->loadBefore* in the hope to simplify the code and keep diff of changes minimal, but in the end it [did not work](https://github.com/zopefoundation/ZODB/pull/323#discussion_r627941838) like that. And also, like I said above, with loadAt it is straightforward to e.g. directly use tid, that lastTransaction returns, while with loadBefore that tid has to be converted to `before = p64(u64(tid)+1)`.

Let me qoute https://github.com/zopefoundation/ZODB/pull/341#issuecomment-973112210 to explain the latter point:

---- 8< ----
@d-maurer, please note how loadAt -> loadBeforeEx renaming made the code somewhat less clear: https://github.com/zopefoundation/ZODB/pull/341/commits/c3820ff48eda59f6efc14bedc1d9cc37b6d59c7d.

There, we are obtaining `base.lastTransaction()` and `changes.lastTransaction()` and then want to load data from storages _at_ that particular states. After at -> before transition, we cannot use that lastTransaction return as is and need to do `p64(u64(·)+1)` all the time around it.

To me it is a common situation - when something determines revision as of which database state it wants to observe data snapshot, and then accesses/loads data _at_ that particular revision. Frequently determining revision is done via `.lastTransaction()`, but it can be also done by other means. With loadAt accessing that snapshot is straightforward, while with loadBefore* we/users are forced to do `p64(u64(·)+1)` dance all the time.

I don't insist on this strongly, but maybe it would make sense to reconsider loadAt -> loadBeforeEx rename and rename `loadBeforeEx` back to -> `loadAt` before it is merged?

This will unify loading semantic with `loadSerial` (which loads _at_ particular revision).
And actually we can also deprecate and eventually remove `loadSerial`, because in practice, usage of `loadSerial` can be almost directly replaced with `loadAt`.
---- 8< ----

So @d-maurer, would you please provide some feedback on this? I understand that you might reject the idea to return to loadAt naming, and probably I will accept and live with that, but it would be a pity not to use a more fitting version. That's why I'm asking for you to please review this once again from scratch.

Kirill

navytux:
( synced to updated https://github.com/zopefoundation/ZODB/pull/323 to resolve conflict on CHANGES.rst )

d-maurer:
Kirill Smelkov wrote at 2022-3-17 09:48 -0700:
> ...
>Dieter, would you please try to review your decision to prefer loadBefore-style for new function compared to loadAt?

I have not advocated for `loadBeforeEx` but to change (incompatibly)
the behavior of `loadBefore` for delete records.
The `loadBeforeEx` has been a compromise, you proposed; I dislike the name
(likely as much as you do).

`loadAt` is a good name -- but ZODB internals use the "before" semantics
and forward and backwards adaptations are necessary to switch
to `loadAt` -- or a rewrite of those ZODB internals.

I am also author and maintainer of `dm.historical` which uses
the "before" semantics (because the ZODB does).
It would not be difficult to switch
to "at" semantics but it is nasty to have code depending
on the ZODB version.


>I understand that you still might be keeping on your preferences, but maybe taking another look afresh might be helpful.

Personally, I do not see the need to change anything.
Your change allows to fix a bug in `DemoStorage` (and other
stacked storages) -- but this bug affects only "delete" records
(which I never have used). I have also never used
the stacking feature of `DemoStorage`.
Thus, form my point of view, this is a bug which does not require fixing --
at least not at the cost of a new load method.
Unlike you, I would not mind that `loadBefore` would change
its behavior for "delete" records in an incompatible way.

navytux:
@d-maurer, thanks for feedback.

> Personally, I do not see the need to change anything.
> ...

To fix the data corruption bug is not the only motivation for me. As said in the description of https://github.com/zopefoundation/ZODB/pull/323 fixing data corruption for DemoStorage is only the first point. The second point is to save work for storages beside FileStorage to save time on every load by not _unneedingly_ retrieving next_serial from disk/database. This change does affect regular - not only deletion - records. And that's why a new method is intoduced to avoid backward compatibility breakage to loadBefore.

----

Regarding ZODB internals - all that is needed to switch them from loadBeforeEx back to loadAt is to semantically revert the following patch: https://github.com/zopefoundation/ZODB/commit/805bf36f4442. I can do that in practically no time, and we already know that after that amendment, the whole amount of changes to ZODB internals of whole #323 will stay almost the same. So in my view ZODB _internals_ are not keeping us to use "before" semantic.

After reading your message about [dm.historical](https://pypi.org/project/dm.historical/) I was initially going to agree that if dm.historical uses "before" semantic via loadBefore, then it makes sense not to break that and keep #323 in current form as is.

But I looked at dm.historical sources and ... as I see it, nothing there uses loadBefore at all.

If I understood it correctly, dm.historical is published only in the form of tarballs on PyPI, So I will refer to relevant parts of dm.historical source code by quoting them here from what I obtained via [dm.historical-2.0.4.tar.gz](https://files.pythonhosted.org/packages/93/c7/cd3d40da4c6791bc64a646fb0417350bda54d00e63eb08324677c84f44e4/dm.historical-2.0.4.tar.gz):

The first place that depends on "before" semantic is Connection class:

```py
class Connection(Connection):
  def __init__(self, baseConnection, timeOrSerial, before=False):
    '''creates a (historical) connection based on *baseConnection* that delivers state at or before *timeOrSerial*.

    *baseConnection* must be an (open) ZODB connection. Its 'db' is used
    as our 'db' as well.

    *timeOrSerial* is either a time ('DateTime' or float/int) or a serial/transactionid ('string').
    '''
```

but it inherits from ZODB.Connection.Connection and passes timeOrSerial to that ZODB.Connection constructor as `before=...`:

```py
    ts = None
    if isinstance(timeOrSerial, bytes): ts = TimeStamp(timeOrSerial)
    if isinstance(timeOrSerial, DateTime): timeOrSerial = timeOrSerial.timeTime()
    if isinstance(timeOrSerial, (int, float)):
      ts = TimeStamp(*(gmtime(timeOrSerial)[:5] + (timeOrSerial%60,)))
    db = baseConnection.db()
    c = super(Connection, self).__init__
    if "before" in getargspec(c).args:
      # this is a modern ZODB which directly supports historical connections
      if not before:
        # we must slightly increment "ts" as we need the state
        #   `at or before` while the connection gives us `before`.
        ts = ts.laterThan(ts)
      c(db, before=to_bytes(ts))
      self.open() 
```

this will _continue to work_ irregardless of whether ZODB internally uses loadBefore, loadBeforeEx or loadAt, because after any form of #323 ZODB.Connection stays fully backward compatible!

----

The other place in dm.historical is `_HistoricalStorageWrapper` class. But it only uses `.history` and `.loadSerial` methods of a given storage. It does not use loadBefore at all! And nothing in dm.historical uses loadBefore _at_ _all_ !

This means that whatever we do - merge #323 with loadBeforeEx, or merge #323 with loadAt - dm.historical will continue to work without any change to its sources.

Does all this maybe change anything in your view, please?


Kirill

d-maurer:
Kirill Smelkov wrote at 2022-3-17 11:35 -0700:
***@***.***, thanks for feedback.
>
>> Personally, I do not see the need to change anything.
>> ...
>
>To fix the data corruption bug is not the only motivation for me. As said in the description of https://github.com/zopefoundation/ZODB/pull/323 fixing data corruption for DemoStorage is only the first point. The second point is to save work for storages beside FileStorage to save time on every load by not _unneedingly_ retrieving next_serial from disk/database. This change does affect regular - not only deletion - records. And that's why a new method is intoduced to avoid backward compatibility breakage to loadBefore.

I have not yet seen a statement about your proposal
from the authors of those storages -- apart from some concerns
regarding `relstorage` (maybe meanwhile resolved).

We have discussed that ZEO uses the additional information
(you want to drop) for its cache consistency. You might have
a solution concept - but it is not yet implemented.


>Regarding ZODB internals - all that is needed to switch them from loadBeforeEx back to loadAt is to semantically revert the following patch: https://github.com/zopefoundation/ZODB/commit/805bf36f4442. I can do that in practically no time, and we already know that after that amendment, the whole amount of changes to ZODB internals of whole #323 will stay almost the same. So in my view ZODB _internals_ are not keeping us to use "before" semantic.

I remember that you had (of course) implemented the
"before -> at" semantic adaptations. They have been the reason
for my initial objection -- the ZODB should either
consistently use "at" or "before" semantics, not switch between
the two. I advocated to stay with "before" semantics.


> ...
>But I looked at dm.historical sources and ... as I see it, nothing there uses loadBefore at all.
> ...
>This means that whatever we do - merge #323 with loadBeforeEx, or merge #323 with loadAt - dm.historical will continue to work without any change to its sources.

You are right: like most of my packages, `dm.historical` is very stable
and it was a long time ago I actively worked on it. I was wrong
to fear that your proposal could break something.


>Does all this maybe change anything in your view, please?

No, it does not.

navytux:
@d-maurer, thanks for feedback.

> I have not yet seen a statement about your proposal from the authors of those storages

Well, when I'm speaking about NEO, I'm meaning to be the author of NEO/go version myself

https://navytux.spb.ru/~kirr/neo

I also contributed to, and have more-or-less thorough understanding of, NEO/py implementation.

> We have discussed that ZEO uses the additional information
> (you want to drop) for its cache consistency. You might have
> a solution concept - but it is not yet implemented.

This is valid point. Without rework of ZEO cache loadBefore cannot be deprecated and dropped.

Will it be accepted if I implement what I had described in https://github.com/zopefoundation/ZODB/pull/323#issuecomment-845917355 and prove that it works well for ZEO and in general?

> I remember that you had (of course) implemented the
> "before -> at" semantic adaptations. They have been the reason
> for my initial objection -- the ZODB should either
> consistently use "at" or "before" semantics, not switch between
> the two. I advocated to stay with "before" semantics.

I agree that consistency is a good goal. But ZODB is already inconsistent: loadBefore works on tid as "before", while e.g. loadSerial accepts tid as "at". And e.g. ZODB.DB - its public interface - accepts both "at" and "before" in its `.open`.

We cannot change the public interface due to backward compatibility. But we can make the internals consistent. Currently they are already inconsistent in between loadBefore and loadSerial. With loadAt and loadSerial everything will be consistently using "at" semantic. Of course, *if* loadBefore gets deprecated, which, once again, depends on ZEO cache rework.

> You are right ...

Thanks for confirming.

> No, it does not.

I'm sorry to hear that, but I understand your reasoning. It is fair to say that loadAt is not generally very useful until loadBefore cannot be deprecated for real because ZEO cache depends on it.

Once again, I'm willing to fix ZEO cache not to depend on `next_serial` sooner or later.

Kirill

d-maurer:
Kirill Smelkov wrote at 2022-3-18 02:06 -0700:
> ...
>> We have discussed that ZEO uses the additional information
>> (you want to drop) for its cache consistency. You might have
>> a solution concept - but it is not yet implemented.
>
>This is valid point. Without rework of ZEO cache loadBefore cannot be deprecated and dropped.
>
>Will it be accepted if I implement what I had described in https://github.com/zopefoundation/ZODB/pull/323#issuecomment-845917355 and prove that it works well for ZEO and in general?

I cannot speak about whether it is accepted or not -- only about
whether I might approve it or not.

The general concept may work and I might approve an implementation.
However, I need to see the concrete implementation (and what it entails)
before my approval.

> ...
>I agree that consistency is a good goal. But ZODB is already inconsistent: loadBefore works on tid as "before", while e.g. loadSerial accepts tid as "at". And e.g. ZODB.DB - its public interface - accepts both "at" and "before" in its `.open`.
>
>
>We cannot change the public interface due to backward compatibility. But we can make the internals consistent. Currently they are already inconsistent in between loadBefore and loadSerial.

You are aware that your "at" (which acutally means "<= at") is
different from the "at" of `loadSerical` (which means "== at").
Otherwise, we would not need a `loadAt` but could use `loadSerial`).

I recognize that the ZODB uses "before" semantics for historical reasons:
formerly, the first invalidation serial after a connection open
defined the snapshot boundary amd the connection read the state
*before* that boundary. Now, the MVCC adapter maintains
information about the last transaction and the connection
loads "at" this transaction boundary.
For historical reasons this "at" is transformed into a "before".
With your `loadAt`,
the "before" must later be retransformed into "at" semantics.
I much dislike this double transformation -- especially, because
the `TimeStamp` does not provide an API for the back transformation.

If we should switch to `Storage.loadAt`, we should rework the
ZODB code to consistently use "at" semantics.

> ...
>Once again, I'm willing to fix ZEO cache not to depend on `next_serial` sooner or later.

`ZEO` and `FileStorage` are the only storages I am currently really interested
in. Thus, my personal decisions depend on good support for them.

navytux:
@d-maurer, thanks for feedback. It is all clear and reasonable. I understand and agree that FileStorage and ZEO should remain with good support, as those storages are historically the most widely used. And that it requires to have the concrete implementation to be able to review it and its consequences. So let's wait for me to do this next step. I do not promise it will be soon, but, hopefully, I should be able to do it.

P.S. and you are right - loadSerial has `==at` semantic while loadAt has `<=at` semantic. However loadSerial can be also replaced with loadAt usage - it is just loadAt(at) + assert that loaded serial == at. So in the end ZODB code could be using at everywhere internally and only loadAt.

PR #342, opened 2021-03-30
Security fix documentation dependencies

icemac:
According to dependabot Pygment versions < 2.7.4 are vulnerable.
I already switched https://readthedocs.org/projects/zodb-docs/ to Python 3.x.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,9 @@
 - Rework ``fsrefs`` script to work significantly faster by optimizing how it does
   IO. See `PR 340 <https://github.com/zopefoundation/ZODB/pull/340>`_.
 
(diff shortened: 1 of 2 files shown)


jamadden (approved):
[The vulnerability](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=985574) is that "By crafting malicious input, an attacker can cause a denial of service."

The use of pygments here is static and completely under our control; it's not used dynamically or exposed to unknown input, e.g., through a web interface. So any risk here seems very low, though not 0. 

This LGTM. Maybe add a comment to the file about why the requirement, and also that the docs can no longer be built with Python 2?

icemac:
@jamadden Agreed. Thank you for reviewing this PR. 😃

PR #343, opened 2021-04-01
Update README.rst

claell:
Fix tiny issue with a remaining colon that was probably not deleted during updates of the README.


jugmac00:
Thank you!

PR #345, opened 2021-04-14
tests: Add test for open vs invalidation race

navytux:
Add test that exercises open vs invalidation race condition that, if
happen, leads to data corruption. We are seeing such race happening on
storage level in ZEO (https://github.com/zopefoundation/ZEO/issues/166),
and previously we've seen it also to happen on Connection level
(https://github.com/zopefoundation/ZODB/issues/290). By adding this test
to be exercised wrt all storages we make sure that all storages stay
free from this race.

And it payed out. Besides catching original problems from
https://github.com/zopefoundation/ZODB/issues/290 and
https://github.com/zopefoundation/ZEO/issues/166 , this test also
discovered a concurrency bug in MVCCMappingStorage:

    Failure in test check_race_open_vs_invalidate (ZODB.tests.testMVCCMappingStorage.MVCCMappingStorageTests)
      File "/usr/lib/python2.7/unittest/case.py", line 329, in run
        testMethod()
      File "/home/kirr/src/wendelin/z/ZODB/src/ZODB/tests/BasicStorage.py", line 492, in check_race_open_vs_invalidate
        self.fail(failure[0])
      File "/usr/lib/python2.7/unittest/case.py", line 410, in fail
        raise self.failureException(msg)
    AssertionError: T1: obj1.value (24)  !=  obj2.value (23)

The problem with MVCCMappingStorage was that instance.poll_invalidations
was correctly taking main_lock with intention to make sure main data is
not mutated during analysis, but instance.tpc_finish and
instance.tpc_abort did _not_ taken main lock, which was leading to
committed data to be propagating into main storage in non-atomic way.

This bug was also observable if both obj1 and obj2 in the added test
were always loaded from the storage (added obj2._p_invalidate after
obj1._p_invalidate).

-> Fix MVCCMappingStorage by correctly locking main MVCCMappingStorage
instance when processing transaction completion.

/cc @d-maurer, @jamadden, @jmuchemb

### DIFF
--- a/src/ZODB/tests/BasicStorage.py
+++ b/src/ZODB/tests/BasicStorage.py
@@ -29,6 +29,7 @@ import threading
 import time
 import zope.interface
 import zope.interface.verify
+from random import randint
 
 from .. import utils
 
@@ -480,3 +481,130 @@ class BasicStorage(object):
 
         if failed.is_set():
             self.fail(failure[0])
+
+
+    # client-server storages like ZEO, NEO and RelStorage allow several storage
+    # clients to be connected to single storage server.
+    #
+    # For client-server storages test subclasses should implement
+    # _new_storage_client to return new storage client that is connected to the
+    # same storage server self._storage is connected to.
+    def _new_storage_client(self):
+        raise NotImplementedError
+
+    # verify storage for race in between load and external invalidations.
+    # https://github.com/zopefoundation/ZEO/issues/155
+    #
+    # This test is similar to check_race_loadopen_vs_local_invalidate but does
+    # not reuse its code because the probability to reproduce external
+    # invalidation bug with only 1 mutator + 1 verifier is low.
+    def check_race_load_vs_external_invalidate(self):
+        # dbopen creates new client storage connection and wraps it with DB.
+        def dbopen():
+            try:
+                zstor = self._new_storage_client()
+            except NotImplementedError:
+                # the test will be skipped from main thread because dbopen is
+                # first used in init on the main thread before any other thread
+                # is spawned.
+                self.skipTest("%s does not implement _new_storage_client" % type(self))
+            return DB(zstor)
+
+        # init initializes the database with two integer objects - obj1/obj2 that are set to 0.
+        def init():
+            db = dbopen()
+
+            transaction.begin()
+            zconn = db.open()
+
+            root = zconn.root()
+            root['obj1'] = MinPO(0)
+            root['obj2'] = MinPO(0)
+
+            transaction.commit()
+            zconn.close()
+
+            db.close()
+
+        # we'll run 8 T workers concurrently. As of 20210416, due to race conditions
+        # in ZEO, it triggers the bug where T sees stale obj2 with obj1.value != obj2.value
+        #
+        # The probability to reproduce the bug is significantly reduced with
+        # decreasing n(workers): almost never with nwork=2 and sometimes with nwork=4.
+        nwork = 8
+
+        # T is a worker that accesses obj1/obj2 in a loop and verifies
+        # `obj1.value == obj2.value` invariant.
+        #
+        # access to obj1 is organized to always trigger loading from zstor.
+        # access to obj2 goes through zconn cache and so verifies whether the cache is not stale.
+        #
+        # Once in a while T tries to modify obj{1,2}.value maintaining the invariant as
+        # test source of changes for other workers.
+        failed = threading.Event()
+        failure = [None] * nwork # [tx] is failure from T(tx)
+        def T(tx, N):
+            db = dbopen()
+
+            def t_():
+                transaction.begin()
+                zconn = db.open()
+
+                root = zconn.root()
+                obj1 = root['obj1']
+                obj2 = root['obj2']
+
+                # obj1 - reload it from zstor
+                # obj2 - get it from zconn cache
+                obj1._p_invalidate()
+
+                # both objects must have the same values
+                i1 = obj1.value
+                i2 = obj2.value
+                if i1 != i2:
+                    #print('FAIL')
+                    failure[tx] = "T%s: obj1.value (%d)  !=  obj2.value (%d)" % (tx, i1, i2)
+                    failed.set()
+
+                # change objects once in a while
+                if randint(0,4) == 0:
+                    #print("T%s: modify" % tx)
+                    obj1.value += 1
+                    obj2.value += 1
+
+                try:
+                    transaction.commit()
+                except POSException.ConflictError:
+                    #print('conflict -> ignore')
+                    transaction.abort()
+
+                zconn.close()
+
+            try:
+                for i in range(N):
+                    #print('T%s.%d' % (tx, i))
+                    t_()
+                    if failed.is_set():
+                        break
+            except:
+                failed.set()
+                raise
+            finally:
+                db.close()
+
+
+        # run the workers concurrently.
+        init()
+
+        N = 100
+        tg = []
+        for x in range(nwork):
+            t = threading.Thread(name='T%d' % x, target=T, args=(x, N))
+            t.start()
+            tg.append(t)
+
+        for t in tg:
+            t.join(60)
+
+        if failed.is_set():
+            self.fail([_ for _ in failure if _])
--- a/src/ZODB/tests/BasicStorage.py
+++ b/src/ZODB/tests/BasicStorage.py
@@ -18,7 +18,8 @@ http://www.zope.org/Documentation/Developer/Models/ZODB/ZODB_Architecture_Storag
 
 All storages should be able to pass these tests.
 """
-from ZODB import POSException
+import transaction
+from ZODB import DB, POSException
 from ZODB.Connection import TransactionMetaData
 from ZODB.tests.MinPO import MinPO
 from ZODB.tests.StorageTestBase import zodb_unpickle, zodb_pickle
@@ -385,3 +386,97 @@ class BasicStorage(object):
         self.assertEqual(results.pop('lastTransaction'), tids[1])
         for m, tid in results.items():
             self.assertEqual(tid, tids[1])
+
+
+    # verify storage/Connection for race in between load/open and local invalidations.
+    # https://github.com/zopefoundation/ZEO/issues/166
+    # https://github.com/zopefoundation/ZODB/issues/290
+    def check_race_loadopen_vs_local_invalidate(self):
+        db = DB(self._storage)
+
+		# init initializes the database with two integer objects - obj1/obj2
+        # that are set to 0.
+        def init():
+            transaction.begin()
+            zconn = db.open()
+
+            root = zconn.root()
+            root['obj1'] = MinPO(0)
+            root['obj2'] = MinPO(0)
+
+            transaction.commit()
+            zconn.close()
+
+        # verify accesses obj1/obj2 and verifies that obj1.value == obj2.value
+        #
+        # access to obj1 is organized to always trigger loading from zstor.
+        # access to obj2 goes through zconn cache and so verifies whether the
+        # cache is not stale.
+        failed = threading.Event()
+        failure = [None]
+        def verify():
+            transaction.begin()
+            zconn = db.open()
+
+            root = zconn.root()
+            obj1 = root['obj1']
+            obj2 = root['obj2']
+
+            # obj1 - reload it from zstor
+            # obj2 - get it from zconn cache
+            obj1._p_invalidate()
+
+            # both objects must have the same values
+            v1 = obj1.value
+            v2 = obj2.value
+            if v1 != v2:
+                failure[0] = "verify: obj1.value (%d)  !=  obj2.value (%d)" % (v1, v2)
+                failed.set()
+
+            transaction.abort() # we did not changed anything; also fails with commit
+            zconn.close()
+
+        # modify changes obj1/obj2 by doing `objX.value += 1`.
+        #
+        # Since both objects start from 0, the invariant that
+        # `obj1.value == obj2.value` is always preserved.
+        def modify():
+            transaction.begin()
+            zconn = db.open()
+
+            root = zconn.root()
+            obj1 = root['obj1']
+            obj2 = root['obj2']
+            obj1.value += 1
+            obj2.value += 1
+            assert obj1.value == obj2.value
+
+            transaction.commit()
+            zconn.close()
+
+        # xrun runs f in a loop until either N iterations, or until failed is set.
+        def xrun(f, N):
+            try:
+                for i in range(N):
+                    #print('%s.%d' % (f.__name__, i))
+                    f()
+                    if failed.is_set():
+                        break
+            except:
+                failed.set()
+                raise
+
+
+        # loop verify and modify concurrently.
+        init()
+
+        N = 500
+        tverify = threading.Thread(name='Tverify', target=xrun, args=(verify, N))
+        tmodify = threading.Thread(name='Tmodify', target=xrun, args=(modify, N))
+        tverify.start()
+        tmodify.start()
+        tverify.join(60)
+        tmodify.join(60)
+
+        if failed.is_set():
+            self.fail(failure[0])
--- a/src/ZODB/tests/BasicStorage.py
+++ b/src/ZODB/tests/BasicStorage.py
@@ -24,6 +24,7 @@ from ZODB.Connection import TransactionMetaData
 from ZODB.tests.MinPO import MinPO
 from ZODB.tests.StorageTestBase import zodb_unpickle, zodb_pickle
 from ZODB.tests.StorageTestBase import ZERO
+from ZODB.tests.util import with_high_concurrency
 
 import threading
 import time
@@ -392,6 +393,7 @@ class BasicStorage(object):
     # verify storage/Connection for race in between load/open and local invalidations.
     # https://github.com/zopefoundation/ZEO/issues/166
     # https://github.com/zopefoundation/ZODB/issues/290
+    @with_high_concurrency
     def check_race_loadopen_vs_local_invalidate(self):
         db = DB(self._storage)
 
@@ -498,6 +500,7 @@ class BasicStorage(object):
     # This test is similar to check_race_loadopen_vs_local_invalidate but does
     # not reuse its code because the probability to reproduce external
     # invalidation bug with only 1 mutator + 1 verifier is low.
+    @with_high_concurrency
     def check_race_load_vs_external_invalidate(self):
         # dbopen creates new client storage connection and wraps it with DB.
         def dbopen():
--- a/src/ZODB/tests/util.py
+++ b/src/ZODB/tests/util.py
@@ -19,6 +19,7 @@ import atexit
 import os
 import persistent
 import re
+import sys
 import tempfile
 import time
 import transaction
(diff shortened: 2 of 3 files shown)


navytux:
CI is failing with:

```
318Version and requirements information containing docutils:
319  Base installation request: 'Sphinx', 'docutils', 'ZODB', 'sphinxcontrib_zopeext', 'j1m.sphinxautozconfig', 'sphinx_rtd_theme'
320  Requirement of sphinx_rtd_theme: docutils<0.17
321  Requirement of j1m.sphinxautozconfig: docutils
322  Requirement of Sphinx: docutils<0.17,>=0.12
323While:
324  Installing sphinx.
325Error: There is a version conflict.
326We already have: docutils 0.17
327but sphinx-rtd-theme 0.5.2 requires 'docutils<0.17'.
328
```

which should be unrelated to my patch.

jmuchemb:
Ouch, that would be the longest ZODB test. I run them for NEO on my laptop:
- before
  ```
       Test Module |  run  | unexpected | expected | skipped |  time    
  -----------------+-------+------------+----------+---------+----------
             Basic |   16  |       .    |      .   |     .   |   1.88s   
          Conflict |    5  |       .    |      .   |     .   |   0.54s   
           History |    1  |       .    |      .   |     .   |   0.15s   
          Iterator |    9  |       .    |      .   |     .   |   1.01s   
                MT |    5  |       .    |      .   |     .   |   5.31s   
              Pack |   14  |       .    |      2   |     .   |  17.49s   
        Persistent |    1  |       .    |      .   |     .   |   0.15s   
          ReadOnly |    2  |       .    |      .   |     .   |   0.30s   
          Revision |    7  |       .    |      .   |     .   |   1.61s   
   Synchronization |    7  |       .    |      .   |     .   |   0.50s   
              Undo |   40  |       .    |      2   |     .   |  11.84s   
              ZODB |   22  |       .    |      1   |     .   |   4.09s   
  -----------------+-------+------------+----------+---------+----------
    neo.tests.zodb |       |            |          |         |          
  -----------------+-------+------------+----------+---------+----------
           Summary |  129  |       .    |      5   |     .   |  44.86s   
  -----------------+-------+------------+----------+---------+----------
  ```
- after
  ```
       Test Module |  run  | unexpected | expected | skipped |  time
  -----------------+-------+------------+----------+---------+----------
             Basic |   17  |       .    |      .   |     .   |  17.29s
          Conflict |    5  |       .    |      .   |     .   |   0.54s
           History |    1  |       .    |      .   |     .   |   0.13s
          Iterator |    9  |       .    |      .   |     .   |   0.99s
                MT |    5  |       .    |      .   |     .   |   5.27s
              Pack |   14  |       .    |      2   |     .   |  17.25s
        Persistent |    1  |       .    |      .   |     .   |   0.17s
          ReadOnly |    2  |       .    |      .   |     .   |   0.27s
          Revision |    7  |       .    |      .   |     .   |   1.65s
   Synchronization |    7  |       .    |      .   |     .   |   0.51s
              Undo |   40  |       .    |      2   |     .   |  11.92s
              ZODB |   22  |       .    |      1   |     .   |   4.04s
  -----------------+-------+------------+----------+---------+----------
    neo.tests.zodb |       |            |          |         |
  -----------------+-------+------------+----------+---------+----------
           Summary |  130  |       .    |      5   |     .   |  60.03s
  -----------------+-------+------------+----------+---------+----------
  ```
So ~15s.

navytux:
We can reduce N somewhat, but the timings highlight that NEO/SQLite is 2x slower than ZEO here. And for ZEO with N=1000 it is only 2 seconds on my machine: (ZEO 5.2.2-1-g3d90ed42 + https://github.com/zopefoundation/ZEO/pull/167; NEO v1.12-77-gde0feb4e):

```console
# stable with 3 repeats
kirr@deco:~/src/wendelin/wendelin.core/lib/tests/testprog$ time WENDELIN_CORE_TEST_DB='<zeo>' ./zopenrace.py
OK

real    0m2,340s
user    0m2,561s
sys     0m0,505s
```

```console
# stable with 3 repeats
kirr@deco:~/src/wendelin/wendelin.core/lib/tests/testprog$ time WENDELIN_CORE_TEST_DB='<neo>' ./zopenrace.py
Using temp directory /tmp/neo_a8sq1P
OK

real    0m4,435s
user    0m4,238s
sys     0m1,026s
```

Reducing N will reduce the probability to catch problems and so will make CI less useful.

I would say that spending even 10 seconds to reliably catch data corruption bugs is worth it.

I will reduce N to 500. On my machine from 2016 (from 5 years ago) this will make the test to run with ZEO for ~ 1s and to run with NEO for ~ 2s.  I believe this is reasonable.

Kirill

navytux:
Travis CI failures are, again, completely unrelated to my patch.

jmuchemb:
> the timings highlight that NEO/SQLite is 2x slower than ZEO here. 

For those who wonder, the ZODB tests in NEO run by default in a single-process mode with a special scheduler. Compared to a normal NEO DB:
- a test DB is slower
- the test has more control over the DB and some operations like setting up the DB are much faster

In normal mode, the ZODB tests of NEO take 100.97s vs 109.18s, so the added test is about twice as fast, which would be about the same as ZEO.

> I will reduce N to 500.

Which which value can you almost always reproduce all bugs that have been found ?

In fact, my previous comment was mainly informative, with little hope that something could be done.

More generally, a "classic" test suite is not really the place for this kind of test. In NEO, we put such test in a stress test suite that we choose for example to run for 1 hour. That would be much more work to do the same in ZODB.

Depending on the value of a useful N, this PR can be fine for now.

navytux:
@jmuchemb, in https://github.com/zopefoundation/ZODB/pull/345#issuecomment-820091765 I ran [zopenrace.py](https://lab.nexedi.com/nexedi/wendelin.core/blob/fe369d32/lib/tests/testprog/zopenrace.py), not ZODB test itself (which is modelled after `zopenrace.py`). That `zopenrace.py` [uses `wendelin.lib.testing.getTestDB`](https://lab.nexedi.com/nexedi/wendelin.core/blob/fe369d32/lib/tests/testprog/zopenrace.py#L109)  to setup test database. With `WENDELIN_CORE_TEST_DB='<neo>'` that `getTestDB` [uses `neo.tests.functional`](https://lab.nexedi.com/nexedi/wendelin.core/blob/fe369d32/lib/testing.py#L289-305) to setup test NEO cluster. Do I read you right that [`neo.tests.functional`](https://lab.nexedi.com/nexedi/neoppod/blob/master/neo/tests/functional/__init__.py) employs that special scheduler instead of regular OS processes? I have always thought that - contrary to [`neo.tests.threaded`](https://lab.nexedi.com/nexedi/neoppod/blob/master/neo/tests/threaded/__init__.py) - `neo.tests.functional` uses regular OS processes because I see `os.fork` in its code and I see forked processes in `ps` output when `zopenrace.py` is being run. I might be missing something, but does your argument really explain the speed difference?

For the reference, here is [how ZEO test cluster is set up](https://lab.nexedi.com/nexedi/wendelin.core/blob/fe369d32/lib/testing.py#L243-286) when `zopenrace.py` is run with `WENDELIN_CORE_TEST_DB='<zeo>'`.

> Which which value can you almost always reproduce all bugs that have been found ?

N=1000. Even 500 is somewhat a compromise, because I rememeber sometimes some tests were failing around iter=700 or 800.

> More generally, a "classic" test suite is not really the place for this kind of test. In NEO, we put such test in a stress test suite that we choose for example to run for 1 hour. That would be much more work to do the same in ZODB.

It is a reasonable approach when it takes non-small time to reproduce problems. But if it just several seconds to reproduce, the place for such tests is in regular testsuite in my view. The synergy of those two approaches could be that such test live in regular testsuite and leverage reasonably small N by default for the run to take seconds. But when the test suite is run with something like `--stress` or `--long` mode, those N are raised up by say 10, 100 or more to exercise problems much more thoroughly.

> Depending on the value of a useful N, this PR can be fine for now.

I'm fine with N=1000; I'm somewhat ok with N=500. In my view the test becomes less useful if N is reduced further.

Kirill

navytux:
I have updated this pull request with another test that exercises "load vs external invalidation" data corruption discovered in https://github.com/zopefoundation/ZEO/issues/155. Quoting https://github.com/zopefoundation/ZODB/pull/345/commits/9601756039921558affd8f0e30d19d1f0bb1d96d :

---- 8< ----
tests: Add test for load vs external invalidation race

For ZEO this data corruption bug was reported at
https://github.com/zopefoundation/ZEO/issues/155 and fixed at
https://github.com/zopefoundation/ZEO/pull/169.

Without that fix the failure shows e.g. as follows when running ZEO test
suite:

    Failure in test check_race_load_vs_external_invalidate (ZEO.tests.testZEO.BlobAdaptedFileStorageTests)
      File "/usr/lib/python2.7/unittest/case.py", line 329, in run 
        testMethod()
      File "/home/kirr/src/wendelin/z/ZODB/src/ZODB/tests/BasicStorage.py", line 621, in check_race_load_vs_external_invalidate
        self.fail([_ for _ in failure if _]) 
      File "/usr/lib/python2.7/unittest/case.py", line 410, in fail
        raise self.failureException(msg)
    AssertionError: ['T1: obj1.value (7)  !=  obj2.value (8)']

Even if added test is somewhat similar to
check_race_loadopen_vs_local_invalidate, it is added anew without trying
to unify code. The reason here is that the probability to catch load vs
external invalidation race is significantly reduced when there are only
1 modify and 1 verify workers. The unification with preserving both
tests semantic would make test for "load vs local invalidate" harder to
follow. Sometimes a little copying is better than trying to unify too 
much.

For the test to work, test infrastructure is amended with
._new_storage_client() method that complements ._storage attribute:
client-server storages like ZEO, NEO and RelStorage allow several
storage clients to be connected to single storage server. For 
client-server storages test subclasses should implement
_new_storage_client to return new storage client that is connected to
the same storage server self._storage is connected to. 

For ZEO ._new_storage_client() is added by https://github.com/zopefoundation/ZEO/pull/170

Other client-server storages can follow to implement ._new_storage_client()
and this way automatically activate this "load vs external invalidation"
test when their testsuite is run.

Contrary to test for "load vs local invalidate" N is set to lower value (100),
because with 8 workers the bug is usually reproduced at not-so-high iteration
number (5-10-20).

/cc @d-maurer, @jamadden, @jmuchemb
---- 8< ----

So now this pull-request, together with https://github.com/zopefoundation/ZEO/pull/170, adds test for all data corruption bugs in ZODB5 stack discovered so far.

navytux:
Travis CI [failures](https://github.com/zopefoundation/ZODB/pull/345/checks?check_run_id=2364320504) - all with error to install sphinx - are, again, completely unrelated to my patch.

d-maurer:
Earlier, we observed that race conditions are much harder to detect in Python 3 than Python 2. The experience from https://github.com/zopefoundation/ZEO/issues/168#issuecomment-821829116 explains this observation: obviously, race condition probability increases with the context switch frequency; by default Python 3 switches contexts every 5 ms compared to 100 interpreter instructions for Python 2 (about 0.01 ms on modern hardware); thus, under normal settings, Python 3 is almost sequential compared to Python 2.

We may want to control the context switch frequency for race condition tests

navytux:
@d-maurer, thanks, this might be good idea. I see corresponding changes on https://github.com/zopefoundation/ZEO/pull/167/files/2c98d943d7e191abd64cdac53876267dffac9c74..4f59226553e651e5cc8597e734a084f63501c16e and I've tried to see whether it indeed makes the test fail with python3. Unfortunately I stilll cannot trigger the test to fail, even with something like

```diff
--- a/lib/tests/testprog/zopenrace.py
+++ b/lib/tests/testprog/zopenrace.py
@@ -106,6 +106,10 @@ def __init__(self, i):
 
 @func
 def main():
+    import sys
+    #sys.setcheckinterval(10)
+    sys.setswitchinterval(5*1e-5)
+
     tdb = getTestDB()
     tdb.setup()
     defer(tdb.teardown)
```

I've tried also `1e-6`, `1e-5` and `1e-4` - the test (zopenrace.py) always pass for me with CPython 3.9.2.

For the reference "100 instructions" on my machine is something in between 1e-6 and 1e-5 seconds.

I also see CPython itself has `sys.setswitchinterval(1e-6)` (the lowest possible value) in tests for threading bugs, so still this might be something we could want to consider.

https://github.com/python/cpython/blob/68ba0c67cac10c2545ea3b62d8b161e5b3594edd/Lib/test/test_concurrent_futures.py#L687-L694
https://github.com/python/cpython/blob/68ba0c67cac10c2545ea3b62d8b161e5b3594edd/Lib/test/_test_multiprocessing.py#L4280-L4295
https://github.com/python/cpython/blob/68ba0c67cac10c2545ea3b62d8b161e5b3594edd/Lib/test/test_importlib/test_threaded_import.py#L261-L273
...

However on python2, if I lower the "check interval" from default 100 to 1 (again the lowest possible value), `zopenrace.py` stops to reproduce problem as well.

We might indeed move race tests to layer (as in your changes), or mark with something lile `@with_high_concurrency` decorator, but it is still not completely clear to me what to do in those settings on python3. And python2 "works" out of the box with defaults.

d-maurer:
Kirill Smelkov wrote at 2021-4-19 11:37 -0700:
> ...
> it is still not completely clear to me what to do in those settings on python3. And python2 "works" out of the box with defaults.

Not yet fully understood race conditions will likely never be
detected reliably by tests because the race happens only under quite
specific conditions. For example, the ZEO#167 race conditions tests
all succeed in my local environment while some fail (reliably) on Travis.
Even when the race condition is fully understood, special instrumentation
may be necessary to enforce that some things happen at precise moments.

IMO we should use heuristics and try to improve it when experience
gives some hints. In ZEO#167, I used the heuristics to emulate
Python 2's context switch frequency in Python 3 because experience
demonstrated that some race condition tests revealed problems in Python 2
but not in Python 3. Another heuristics could be to learn from
Python race condition tests. I do not know what is better.

navytux:
@d-maurer, thanks. I have marked my race-reproducing tests with `@with_high_concurrency` decorator that makes sure to run wrapped function with appropriate `sys.set{switch,check}interval`. Details are here: https://github.com/zopefoundation/ZODB/pull/345/commits/acace757bfb449be5e2bcb20c1bade4e834a93fd.

Would you please have a look?
I believe this PR should be ready to go in. And even if later we find something to improve, we can improve that as a next follow-up step.

Thanks beforehand,
Kirill

src/ZODB/tests/util.py
```diff
@@ -338,3 +339,41 @@ def testSetUp(self):
...
+def with_high_concurrency(f):
+    """
+    with_high_concurrency decorates f to run with high frequency of thread context switches.
+
+    It is useful for tests that try to probabilistically reproduce race
+    condition scenarios.
+    """
+    @functools.wraps(f)
+    def _(*argv, **kw):
+        if six.PY3:
```
jmuchemb, line 353: Can you move to an upper level so we can have:
```py
previous = sys.getswitchinterval()
try:
    sys.setswitchinterval(5e-6)
    return f(*argv, **kw)
finally:
    sys.setswitchinterval(previous)
```
and
```py
previous = sys.getcheckinterval()
try:
    sys.setcheckinterval(100)
    return f(*argv, **kw)
finally:
    sys.setcheckinterval(previous)
```

src/ZODB/tests/util.py
```diff
@@ -15,6 +15,7 @@
 """
 from ZODB.MappingStorage import DB
 
+import sys
```
jmuchemb, line 18: Can you move between `re` & `tempfile` so it remains somewhat sorted?

navytux:
CI is ok, modulo [already seen](https://github.com/zopefoundation/ZODB/pull/345#issuecomment-821329124) unrelated "version conflict on install sphinx" [kind of problems](https://travis-ci.com/github/zopefoundation/ZODB/jobs/499760337).

src/ZODB/tests/BasicStorage.py
```diff
@@ -385,3 +388,237 @@ def lastInvalidations():
...
+
+            try:
+                for i in range(N):
+                    #print('modify.%d' % i)
+                    _modify()
+                    if failed.is_set():
+                        break
+            except:
+                failed.set()
+                raise
```
jmuchemb, line 481: I prefer:
1. move _modify/_verify out of modify/verify
2. merge modify & verify into a single function (you could still have a nested function with code that is common to _modify/_verify)
3. rename _modify/_verify to modify/verify
4. make threading.Thread pass an extra argument that is either modify or verify
5. for the commented print, you can use `__name__` of the passed function

On the other side, no need for threading.Thread to pass N.

src/ZODB/tests/BasicStorage.py
```diff
@@ -385,3 +388,237 @@ def lastInvalidations():
...
+        # in ZEO, it triggers the bug where T sees stale obj2 with obj1.value != obj2.value
+        #
+        # The probability to reproduce the bug is significantly reduced with
+        # decreasing n(workers): almost never with nwork=2 and sometimes with nwork=4.
+        nwork = 8
+        init()
+
+        N = 100
+        tg = []
+        failure = [None]*nwork
```
jmuchemb, line 614: missing space around `*`

src/ZODB/tests/BasicStorage.py
```diff
@@ -385,3 +388,237 @@ def lastInvalidations():
...
+        for x in range(nwork):
+            t = threading.Thread(name='T%d' % x, target=T, args=(x, N))
+            t.start()
+            tg.append(t)
```
jmuchemb, line 618: what about
```py
tg = [threading.Thread(name='T%d' % x, target=T, args=(x, N))
      for x in range(nwork)]
for t in tg:
    t.start()

**Review by d-maurer — APPROVED** [review]

**Comment by d-maurer** [comment]
To get rid of the `Sphinx` problem, you could try to add in `buildout.cfg` under section `versions:python2`: `docutils = < 0.17`

**Comment by navytux** [comment]
> To get rid of the `Sphinx` problem, you could try to add in `buildout.cfg` under section `versions:python2`: `docutils = < 0.17`

Thanks, @d-maurer. I've created https://github.com/zopefoundation/ZODB/pull/346 to address this.

File: src/ZODB/tests/util.py (hunk 2 above)
Comment (navytux) [reply src/ZODB/tests/util.py:18]: ok

File: src/ZODB/tests/BasicStorage.py (hunk 3 above)
Comment (navytux) [reply src/ZODB/tests/BasicStorage.py:481]: ok, let's have it this way

> On the other side, no need for threading.Thread to pass N.

except this - I prefer to pass parameters explicitly instead of implicitly referring to shared globals.

￼

File: src/ZODB/tests/BasicStorage.py (hunk 4 above)
Comment (navytux) [reply src/ZODB/tests/BasicStorage.py:614]: ok for both - I'm moving nwork=8 up and now there is no duplicate for failure definition.

File: src/ZODB/tests/BasicStorage.py (hunk 5 above)
Comment (navytux) [reply src/ZODB/tests/BasicStorage.py:618]: it is a bit less clear to me, so I prefer to leave my version of this code.

File: src/ZODB/tests/util.py (hunk 1 above)
Comment (navytux) [reply src/ZODB/tests/util.py:353]: It makes the code less clear to me. I prefer to keep my original version.

**Comment by navytux** [comment]
Changes updated according to review feedback. Unless https://github.com/zopefoundation/ZODB/pull/346 is merged some CI will fail, so if ok let's merge #346 first and then I will incorporate that into hereby PR.

**Comment by navytux** [comment]
https://github.com/zopefoundation/ZODB/pull/346 was merged. I synced changes in this pull-request with master to get to green CI state. Let's see how it goes this time.

File: src/ZODB/tests/util.py (hunk 1 above)
Comment (jmuchemb) [reply src/ZODB/tests/util.py:353]: One could argue about the duplication of the try/return/finally lines, even if for me it's so small that I'd call it "overfactoring". But less clear? There's nothing more pythonic than the following pattern:
```
save state
try:
    mess up (and ideally, do it after the try - see PEP 419)
    ...
finally:
    restore
```
No need to "name" the restore part by moving it a function.

When we drop support for Python 2, that's anyway how it would be rewritten.

navytux: > One could argue about the duplication of the try/return/finally lines

Yes, that duplication of f call sites makes it less clear in my view.

> When we drop support for Python 2, that's anyway how it would be rewritten.

So let's rewrite it that way _if_ | _when_ Python 2 support is dropped. And until it is not dropped, let's keep the code clearly structured.

src/ZODB/tests/BasicStorage.py
jmuchemb: Besides style (here, I won't insist), I wonder it makes a difference to minimize the difference of start time between threads, and if it would have an impact to reproduce failure. Or maybe it's completely negligible.

navytux: I've tried this change and run the tests on unfixed ZEO. The failure is reproducible with both versions. The iteration number at which the failue is reproduced is scattered in between 1-5-6-30-... in both versions. So I believe this change is completely orthogonal to test semantic.

src/ZODB/tests/BasicStorage.py
jmuchemb: Since you capitalized it, it's semantically not a variable but a constant.

navytux: It's just a major parameter.

navytux:
CI is green.

navytux:
I believe it is time to merge this.

navytux:
( patches rebased to master with fixups squashed into corresponding commits )

PR #351, opened 2021-08-25
Fix TypeError for fsoids

### LINKED ISSUE #350: fsoids script fails reporting the shortened tid description on Python 3
## 
fsoids script fails reporting the shortened tid description on Python 3

### What I did:

I ran the fsoids script passing one oid:
```
$ ./bin/fsoids components/zeo/var/filestorage/Data.fs 18104
```

### What I expect to happen:

I expected this output:
```
$ ./bin/fsoids components/zeo/var/filestorage/Data.fs 18104
oid 0x46b8 ZODB.blob.Blob 2 revisions
    tid 0x03e1ec746d9d5a77 offset=32500086 2021-08-06 13:40:25.690936
        tid user=b' ale'
        tid description=b'/Plone/workspaces/open ... 787f6b6dfe6f.jpeg/index_html'
        new revision ZODB.blob.Blob at 32501238
        new revision ZODB.blob.Blob at 32501305
        referenced by 0x46b7 plone.namedfile.file.NamedBlobImage at 32500855
```

### What actually happened:

```
oid 0x46b8 ZODB.blob.Blob 2 revisions
    tid 0x03e1ec746d9d5a77 offset=32500086 2021-08-06 13:40:25.690936
        tid user=b' ale'
  File "./bin/fsoids", line 319, in <module>
    sys.exit(ZODB.scripts.fsoids.main())
  File "/home/ale/.buildout/eggs/ZODB-5.6.0-py3.8.egg/ZODB/scripts/fsoids.py", line 76, in main
    c.report()
  File "/home/ale/.buildout/eggs/ZODB-5.6.0-py3.8.egg/ZODB/FileStorage/fsoids.py", line 120, in report
    print("        tid description=%r" % shorten(description))
  File "/home/ale/.buildout/eggs/ZODB-5.6.0-py3.8.egg/ZODB/FileStorage/fsoids.py", line 32, in shorten
    return s[:nleading] + " ... " + s[-ntrailing:]
TypeError: can't concat str to bytes
```

### What version of Python and Zope/Addons I am using:

- Python 3.8
- ZODB-5.6.0
- Zope-4.6.3
- Plone-5.2.5

ale-rt:
Fix ``TypeError: can't concat str to bytes`` when running fsoids.py script with Python 3.

Closes #350

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@
 5.6.1 (unreleased)
 ==================
 
+- Fix ``TypeError: can't concat str to bytes`` when running fsoids.py script with Python 3.
+  See `issue 350 <https://github.com/zopefoundation/ZODB/issues/350>`_.
+
 - Fix UnboundLocalError when running fsoids.py script.
   See `issue 285 <https://github.com/zopefoundation/ZODB/issues/285>`_.
 
--- a/src/ZODB/FileStorage/fsoids.py
+++ b/src/ZODB/FileStorage/fsoids.py
@@ -29,7 +29,11 @@ def shorten(s, size=50):
     navail = size - 5
     nleading = navail // 2
     ntrailing = size - nleading
-    return s[:nleading] + " ... " + s[-ntrailing:]
+    if isinstance(s, bytes):
+        sep = b" ... "
+    else:
+        sep = " ... "
+    return s[:nleading] + sep + s[-ntrailing:]
 
 class Tracer(object):
     """Trace all occurrences of a set of oids in a FileStorage.


ale-rt:
I made the patch in such a way that the shorten function can handle bytes as well as text.
The code we have now on master actually says we should deal with strings:
https://github.com/zopefoundation/ZODB/blob/1fb097b41cae4ca863c8a3664414c9ec0e204393/src/ZODB/FileStorage/fsoids.py#L24-L32
One other possibility is to enforce the shorten function to refuse any output that is not a native string or to convert it to a native string, but this appears to me that increases the change of breaking some other stuff...

Suggestions are welcome.

Once it is clear what to do I can even add some more tests.

CHANGES.rst
```diff
@@ -5,6 +5,9 @@
 5.6.1 (unreleased)
 ==================
 
+- Fix ``TypeError: can't concat str to bytes`` when running fsoids.py script with Python 3.
```
d-maurer, line 8: Ok

d-maurer (approved):

PR #355, opened 2021-10-03
fsdump/fsstats improvements

### LINKED ISSUE #354: `fsstats` no longer matches output of `fsdump`
ZODB==5.6.0

`scripts.fsstats` is supposed to derive statistics from the output of `FileStorage.fsdump`. However, it no longer understands this output.

Transaction records produced by `fsdump` look like
```
Trans #00000 tid=038e1d5f01b145cc time=2011-05-05 07:59:00.396673 offset=70
    status='p' user='' description='Created Zope Application\n\nAdded temp_folder'
```
and data records like
```
  data #00000 oid=0000000000000006 size=141 class=Products.ZODBMountPoint.MountedObject.MountedObject
```
but `fsstats` tries to match transaction records with re `"tid=([0-9a-f]+).*size=(\d+)"` (i.e. it expects `size` information after `tid`) and data records with re `"oid=([0-9a-f]+) class=(\S+) size=(\d+)"` (i.e. it expects `size` and `class` interchanged).

d-maurer:
Fixes #354.

In details:
 * adds again transaction size information to the output of `fsdump`
 * makes `fsdump` callable via Python's `-m` command line option
 * adapts `fsstats` to `fsdump`'s changed order of `size` and `class` fields in the object information

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,12 @@
 5.6.1 (unreleased)
 ==================
 
+- Readd transaction size information to ``fsdump`` output;
+  adapt `fsstats` to ``fsdump``'s exchanged order for ``size`` and ``class``
+  information in data records;
+  (fixes `#354 <https://github.com/zopefoundation/ZODB/issues/354>_).
+  Make ``fsdump`` callable via Python's ``-m`` command line option.
+
 - Fix UnboundLocalError when running fsoids.py script.
   See `issue 285 <https://github.com/zopefoundation/ZODB/issues/285>`_.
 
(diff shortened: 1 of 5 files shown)


icemac (approved):
LGTM.

Do you you see a chance to write a smoke test, checking that `fsstats` can read the output of `fsdmp` so this API does not break again the next time someone works on `fsdump`?

d-maurer:
> Do you you see a chance to write a smoke test, checking that `fsstats` can read the output of `fsdmp` so this API does not break again the next time someone works on `fsdump`?

It has been a good idea to add tests: `fsstats` still contained several Python 3 incompatibilities.

PR #376, opened 2023-02-03
Fix `racetest` problems

d-maurer:
This PR fixes problems with the `racetest`s observed during work on "https://github.com/zopefoundation/ZEO/pull/225".

It uses daemon (rather than normal) threads to avoid deadlock on shutdown when one of the test threads has not finished. It treats the failure for a test thread to finish as a test failure.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@
 5.8.1 (unreleased)
 ==================
 
+- Fix ``racetest`` problems.
+  For details see `#376 <https://github.com/zopefoundation/ZODB/pull/376>`_.
+
 
 5.8.0 (2022-11-09)
 ==================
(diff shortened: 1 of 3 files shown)


dataflake:
@d-maurer I am currently fixing the GHA configuration-related errors and deprecation warnings on `master` and will let you know when I am done, then you can merge the changes here.

dataflake:
If you merge in https://github.com/zopefoundation/ZODB/commit/9ffc59e5bd3eb68d8e43f79265e45d245b832f76 you should get the py35/py36 tests back and get rid of all the deprecation warnings.

d-maurer:
Jens Vagelpohl wrote at 2023-2-3 03:26 -0800:
>If you merge in https://github.com/zopefoundation/ZODB/commit/9ffc59e5bd3eb68d8e43f79265e45d245b832f76 you should get the py35/py36 tests back and get rid of all the deprecation warnings.

Thank you for your help!

navytux (approved):
Thanks, Dieter. I agree it is better to explicitly fail if thread join does not succeed.

PR #381, opened 2023-05-03
fix(repozo) `--with-verify` argument

templiert:
The argument `--with-verification` for `--recover` does not work.

"verify" is used 39 times in the script, and "verification" is used only 2 times.
I think the two occurrences of `--with-verification` should be `--with-verify`. 
The script then works and the doc is correct.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,9 @@
 - Fix ``racetest`` problems.
   For details see `#376 <https://github.com/zopefoundation/ZODB/pull/376>`_.
 
+- Fix ``--with-verify`` argument in script repozo ``--recover``.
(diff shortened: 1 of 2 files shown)


d-maurer (approved):
Thank you for the fix!

I suggest, you add a corresponding `CHANGES.rst` entry.

templiert:
I merged 2 commits for the fix and for the edit to changes.rst using a [soft reset](https://stackoverflow.com/a/24690646/10832217).
The single commit with everything is now d4ba38123ec2a25a0fb083f9a1d423c86baa10c4.

icemac:
@templiert Thank you for your PR it will be automatically merged after the CI checks succeed.

PR #382, opened 2023-05-24
Remove unused ZODB.tests.warnhook

navytux:
The last usage of WarningsHook was removed in 2011 in f1b04dd8 (Lots of code cleanups.). And when we need to capture warnings there is `warnings.catch_warnings` that is available out of the box from stdlib on both py2 and py3.


d-maurer (approved):

navytux:
Thanks, Dieter.

PR #390, opened 2023-08-10
Update blob.py

### LINKED ISSUE #257: zodb install on termux linux for Android doesn't work due to no os.link

BandungBB:
This fix works for any installation of ZODB on unspecified platforms like Android.  It is proposed for the termux version of ZODB. 

While a pip install zodb within the termux distro for Android completes without incident, it fails to run since any attempts to import ZODB within termux's python fails.  This is because their python's os.link has been disabled for some cross platform reason related to Windows.

The only ZODB library code which uses os.link is in ZODB.blob.py where it is selected as the value for the link_or_copy name on Posix systems.  On Windows systems, it is mapped to shutil.copy.  

This pull request extends the mapping of link_or_copy to shutil.copy for Android, not by attempting to identify the platform as Android, but by attempting a try:except snippet.

Closes #257.

### DIFF
--- a/src/ZODB/blob.py
+++ b/src/ZODB/blob.py
@@ -946,7 +946,7 @@ if sys.platform == 'win32':
 else:
     remove_committed = os.remove
     remove_committed_dir = shutil.rmtree
-    
+
     try:
         link_or_copy = os.link
     except AttributeError:  # pragma: no cover


src/ZODB/blob.py
```diff
@@ -942,8 +942,10 @@ def remove_committed_dir(path):
                 filename = os.path.join(dirpath, filename)
                 remove_committed(filename)
         shutil.rmtree(path)
-
-    link_or_copy = shutil.copy
+    try:
+        link_or_copy = shutil.copy
+    except AttributeError:    #pragma: no cover
+        link_or_copy = shutil.copy
```
tseaver, line 948: First, this cannot be the correct spot for the `try: .. except: ...` bit: it is inside the `if sys.platform == "win32":` block.  The correct location should be below (line 950).

Second, the two parts of the `try: ... except: ...` test must not have the same content.  Line 950 below should be changed to:

```python
    try:
        link_or_copy = os.link
    except AttributeError:  #pragma: no cover
        link_or_copy = shutil.copy
```

tseaver (changes requested):

src/ZODB/blob.py
BandungBB: Ooops.  You are correct. The changes were meant for line 950

src/ZODB/blob.py
```diff
@@ -942,13 +942,14 @@ def remove_committed_dir(path):
                 filename = os.path.join(dirpath, filename)
                 remove_committed(filename)
         shutil.rmtree(path)
-
-    link_or_copy = shutil.copy
+        link_or_copy = shutil.copy
```
tseaver, line 945: This assignment **must** be at the same indentation level as the function definitions:  it is not part of the function.

```suggestion
    link_or_copy = shutil.copy
```

src/ZODB/blob.py
```diff
@@ -942,13 +942,14 @@ def remove_committed_dir(path):
...
 else:
     remove_committed = os.remove
     remove_committed_dir = shutil.rmtree
-    link_or_copy = os.link
-
+    try:
+        link_or_copy = os.link
+    except AttributeError:    #pragma: no cover
+        link_or_copy = shutil.copy
```
tseaver, line 952: Please add a comment indicating why we fall back to `shutil.copy` here, e.g.:
```suggestion
    except AttributeError:    #pragma: no cover
        # FBO termux on Android.
        #  See https://github.com/zopefoundation/ZODB/issues/257
        link_or_copy = shutil.copy
```

tseaver (changes requested):

dataflake:
Please take a look at the GitHub Actions test run. There's some linting issues and this PR breaks all Windows tests.

dataflake:
There's still a linter failure:

```
lint: commands[1]> flake8 src setup.py
src/ZODB/blob.py:945:1: W293 blank line contains whitespace
src/ZODB/blob.py:954:64: W291 trailing whitespace
```

src/ZODB/blob.py
```diff
@@ -942,12 +942,17 @@ def remove_committed_dir(path):
                 filename = os.path.join(dirpath, filename)
                 remove_committed(filename)
         shutil.rmtree(path)
-
+
```
icemac, line 945:
```suggestion

```

src/ZODB/blob.py
```diff
@@ -942,12 +942,16 @@ def remove_committed_dir(path):
                 filename = os.path.join(dirpath, filename)
                 remove_committed(filename)
         shutil.rmtree(path)
-
     link_or_copy = shutil.copy
```
tseaver, line 945: One last niggle:  the empty line separates a function definition from the assignment, so please restore it.

tseaver (changes requested):

BandungBB:
Why would a 5 line fix that is nothing more than a try/except wrapper around the setting of a variable, trigger a Sphinx extension loading error like this one?

Could not import extension sphinxcontrib.zopeext.autointerface (exception: cannot import name 'ObjectMembers' from 'sphinx.ext.autodoc' (/home/runner/work/ZODB/ZODB/.tox/docs/lib/python3.9/site-packages/sphinx/ext/autodoc/__init__.py))

What if I were to destroy my fork of zodb , create another fork, create a pull request on the new fork which includes the recommended change as one clean commit? Could it be that the fork is the problem?

d-maurer:
Bandung wrote at 2023-8-26 20:36 -0700:
>Why would a 5 line fix that is nothing more than a try/except wrapper around the setting of a variable, trigger a Sphinx extension loading error like this one?
>
>Could not import extension sphinxcontrib.zopeext.autointerface (exception: cannot import name 'ObjectMembers' from 'sphinx.ext.autodoc' (/home/runner/work/ZODB/ZODB/.tox/docs/lib/python3.9/site-packages/sphinx/ext/autodoc/__init__.py))

This looks like a version mismatch problem.
A missing "version pin" is the main cause for such problems.
The missing version pin can result in the installation of different
versions in different environments.

Without a version pin, incompatibilities between a newer and an
older version occasionally let newer test runs fail that
have succeeded earlier.


The test failure you observe has nothing to do with your change.
It affects only the documentation generation. The cause is likely
an incompatibility of a newer `sphinx` version (the documentation
generation tool).

When I hit such a case and do not see an obvious fix, I (mostly) ignore
the error, write a PR comment for people more familiar with
the respective failure area and then continue in the normal PR process.

PR #392, opened 2023-10-03
fix sorting dict items

templiert:
`dict_items` instance does not have a `sort` method


dataflake (approved):

icemac:
Thank you for this PR. 😃

PR #396, opened 2024-02-01
repozo: exit with non-zero code in case of verification failure

perrinjerome:
When verification failed, repozo prints a message on standard error, but the program always exits with a return code indicating a success. In case of error it's more natural to exit with an error return code.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@
 
 - Fix sorting issue in ``scripts/space.py``.
 
+- Fix exit code of ``repozo`` script in case of verification error.
+  For details see `#396 <https://github.com/zopefoundation/ZODB/pull/396>`_.
 
 5.8.1 (2023-07-18)
 ==================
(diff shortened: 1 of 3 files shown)


perrinjerome:
This is a trivial change, the script is changed to raise a `VerificationFail` that is catched here and used to set the return code:

https://github.com/zopefoundation/ZODB/blob/ff6c320512096704d5950a46515a1f80e662e010/src/ZODB/scripts/repozo.py#L827-L828

quoting `sys.exit` doc https://docs.python.org/3.13/library/sys.html#sys.exit

> any other object is printed to [stderr](https://docs.python.org/3.13/library/sys.html#sys.stderr) and results in an exit code of 1. In particular, sys.exit("some error message") is a quick way to exit a program when an error occurs.

One visible change is that this only outputs the first error, but every errors are fatal so I was thinking it's not really important to have all error and this made the code a little simpler.

icemac (approved):

perrinjerome:
Thank you @icemac

PR #397, opened 2024-02-14
Fix behavior of ``FileStorage.restore`` when provided wrong ``prev_txn``

perrinjerome:
See discussions from https://github.com/zopefoundation/ZODB/pull/395#discussion_r1474127300 about removing ` + 8 ` from `tend`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,9 @@
 
 - Fix sorting issue in ``scripts/space.py``.
 
+- Fix behavior of ``FileStorage.restore`` when provided wrong ``prev_txn``.
+  For details see `#397 <https://github.com/zopefoundation/ZODB/pull/397>`_.
+
 
 5.8.1 (2023-07-18)
 ==================
--- a/src/ZODB/FileStorage/FileStorage.py
+++ b/src/ZODB/FileStorage/FileStorage.py
@@ -1048,6 +1048,8 @@ class FileStorage(
             tid = decodebytes(transaction_id + b'\n')
             assert len(tid) == 8
             tpos = self._txn_find(tid, 1)
+            if not tpos:
+                raise UndoError("Invalid transaction id")
             tindex = self._txn_undo_write(tpos)
             self._tindex.update(tindex)
             return self._tid, tindex.keys()
@@ -1066,7 +1068,7 @@ class FileStorage(
                 # check the status field of the transaction header
                 if h[16] == b'p':
                     break
-        raise UndoError("Invalid transaction id")
+        return None
 
     def _txn_undo_write(self, tpos):
         # a helper function to write the data records for transactional undo
--- a/src/ZODB/FileStorage/FileStorage.py
+++ b/src/ZODB/FileStorage/FileStorage.py
@@ -672,7 +672,7 @@ class FileStorage(
         tid, tl, status, ul, dl, el = unpack(TRANS_HDR, h)
         status = as_text(status)
         self._file.read(ul + dl + el)
-        tend = tpos + tl + 8
+        tend = tpos + tl
         pos = self._file.tell()
         while pos < tend:
             h = self._read_data_header(pos)
--- a/src/ZODB/tests/testFileStorage.py
+++ b/src/ZODB/tests/testFileStorage.py
@@ -259,6 +259,30 @@ class FileStorageTests(
         self._storage.tpc_finish(t)
         self.assertEqual(self._storage.load(oid)[0], b'data1b')
 
+    def testRestorePrevTxnWithoutOid(self):
+        t = TransactionMetaData()
+        self._storage.tpc_begin(t)
+        self._storage.store(b'\1' * 8, b'\0' * 8, b'data1', b'', t)
+        self._storage.tpc_vote(t)
+        tid1 = self._storage.tpc_finish(t)
+
+        # this transaction is also here to detect problems if restore reads
+        # below the end of tid1
+        t = TransactionMetaData()
+        self._storage.tpc_begin(t)
+        self._storage.store(b'\2' * 8, b'\0' * 8, b'data2a', b'', t)
+        self._storage.tpc_vote(t)
+        self._storage.tpc_finish(t)
+
+        # restore with a transaction not containing the oid
+        t = TransactionMetaData()
+        self._storage.tpc_begin(t)
+        self._storage.restore(b'\2' * 8, b'\0' * 8, b'data2b', b'', tid1, t)
+        self._storage.tpc_vote(t)
+        self._storage.tpc_finish(t)
+        self.assertEqual(self._storage.load(b'\1' * 8)[0], b'data1')
+        self.assertEqual(self._storage.load(b'\2' * 8)[0], b'data2b')
+
     def testCorruptionInPack(self):
         # This sets up a corrupt .fs file, with a redundant transaction
         # length mismatch.  The implementation of pack in many releases of
--- a/src/ZODB/tests/testFileStorage.py
+++ b/src/ZODB/tests/testFileStorage.py
@@ -242,6 +242,23 @@ class FileStorageTests(
         # Before ZODB 3.2.6, this failed, with ._oid == z64.
         self.assertEqual(self._storage._oid, giant_oid)
 
+    def testRestorePrevTxnNotExists(self):
+        t = TransactionMetaData()
+        oid = b'\1' * 8
+        self._storage.tpc_begin(t)
+        self._storage.store(oid, b'\0' * 8, b'data1a', b'', t)
+        self._storage.tpc_vote(t)
+        self._storage.tpc_finish(t)
+
+        # restore with a transaction not existing in the storage
+        tid_not_exist = b'\0\0\0\0\0\1\2\3'
+        t = TransactionMetaData()
+        self._storage.tpc_begin(t)
+        self._storage.restore(oid, b'\0' * 8, b'data1b', b'', tid_not_exist, t)
+        self._storage.tpc_vote(t)
+        self._storage.tpc_finish(t)
+        self.assertEqual(self._storage.load(oid)[0], b'data1b')
+
     def testCorruptionInPack(self):
         # This sets up a corrupt .fs file, with a redundant transaction
         # length mismatch.  The implementation of pack in many releases of


perrinjerome:
I'm not sure what should be the behavior if `restore`  is called with a `prev_txn` which does not contain oid.

The docs are :

https://github.com/zopefoundation/ZODB/blob/6f8878124fa8bf1cc1932a8f67da8d15b653897c/src/ZODB/interfaces.py#L978-L981

https://github.com/zopefoundation/ZODB/blob/6f8878124fa8bf1cc1932a8f67da8d15b653897c/src/ZODB/FileStorage/FileStorage.py#L715-L722


If `prev_txn` does not exist, then it can be ignored. This was not the case so I made a first commit to change this.

But if `prev_txn` exists but does not contain `oid` it might be an error case, in this version it's also ignored.

d-maurer:
Jérome Perrin wrote at 2024-2-14 07:20 -0800:
>I'm not sure what should be the behavior if `restore`  is called with a `prev_txn` which does not contain oid.

I have looked at the code:
The only purpose of `prev_txn` is to determine `prev_pos`.
If this succeeds, we know that we find *data* at `prev_pos`
and therefore we can write a backpointer to `prev_pos` instead
of writing *data* ourself.
This means, `prev_txn` is there only as an optimization.
If it does not contain the correct data, `prep_pos` will be `0`
and the optimazation will not be taken.

perrinjerome:
Thank you for the feedback and sorry for being slow to reply here.

That was also my feeling that restoring a transaction with a `prev_txn` not containing the referenced `oid` is not supposed to happen and could be changed into a log, warning or error. I studied this more and it's a bit more complex.

<details>
<summary>"simple" patch to change to an error</summary>

```diff
diff --git a/src/ZODB/FileStorage/FileStorage.py b/src/ZODB/FileStorage/FileStorage.py
index 5c76000a..869ae8ea 100644
--- a/src/ZODB/FileStorage/FileStorage.py
+++ b/src/ZODB/FileStorage/FileStorage.py
@@ -733,6 +733,11 @@ class FileStorage(
                 prev_txn_pos = self._txn_find(prev_txn, 0)
                 if prev_txn_pos:
                     prev_pos = self._data_find(prev_txn_pos, oid, data)
+                    if not prev_pos:
+                        panic(
+                            "oid %s not found in prev_txn %s",
+                            utils.oid_repr(oid),
+                            utils.tid_repr(prev_txn))
             old = self._index_get(oid, 0)
             # Calculate the file position in the temporary file
             here = self._pos + self._tfile.tell() + self._thl
diff --git a/src/ZODB/tests/testFileStorage.py b/src/ZODB/tests/testFileStorage.py
index 5f545e52..fb21b6d4 100644
--- a/src/ZODB/tests/testFileStorage.py
+++ b/src/ZODB/tests/testFileStorage.py
@@ -260,6 +260,7 @@ class FileStorageTests(
         self.assertEqual(self._storage.load(oid)[0], b'data1b')
 
     def testRestorePrevTxnWithoutOid(self):
+        from ZODB.FileStorage.FileStorage import CorruptedTransactionError
         t = TransactionMetaData()
         self._storage.tpc_begin(t)
         self._storage.store(b'\1' * 8, b'\0' * 8, b'data1', b'', t)
@@ -274,14 +275,13 @@ class FileStorageTests(
         self._storage.tpc_vote(t)
         self._storage.tpc_finish(t)
 
-        # restore with a transaction not containing the oid
+        # restore with a transaction not containing the oid is an error
         t = TransactionMetaData()
         self._storage.tpc_begin(t)
-        self._storage.restore(b'\2' * 8, b'\0' * 8, b'data2b', b'', tid1, t)
-        self._storage.tpc_vote(t)
-        self._storage.tpc_finish(t)
-        self.assertEqual(self._storage.load(b'\1' * 8)[0], b'data1')
-        self.assertEqual(self._storage.load(b'\2' * 8)[0], b'data2b')
+        with self.assertRaisesRegex(
+                CorruptedTransactionError,
+                'oid 0x0202020202020202 not found in prev_txn.*'):
+            self._storage.restore(b'\2' * 8, b'\0' * 8, b'data2b', b'', tid1, t)
 
     def testCorruptionInPack(self):
         # This sets up a corrupt .fs file, with a redundant transaction

```
</details>

I tried to change into an error and there is a legitimate case here in https://github.com/zopefoundation/ZODB/blob/6f8878124fa8bf1cc1932a8f67da8d15b653897c/src/ZODB/tests/RecoveryStorage.py#L41-L69

If this becomes an error, that test fails with an error like this:

```
Error in test testRestoreAcrossPack (ZODB.tests.testFileStorage.FileStorageHexRecoveryTest)
  File "/usr/local/python/3.10.13/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/local/python/3.10.13/lib/python3.10/unittest/case.py", line 591, in run
    self._callTestMethod(testMethod)
  File "/usr/local/python/3.10.13/lib/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/workspaces/ZODB/src/ZODB/tests/RecoveryStorage.py", line 66, in testRestoreAcrossPack
    self._dst.restore(r.oid, r.tid, r.data, '', r.data_txn,
  File "/workspaces/ZODB/src/ZODB/tests/hexstorage.py", line 82, in restore
    return self.base.restore(
  File "/workspaces/ZODB/src/ZODB/FileStorage/FileStorage.py", line 738, in restore
    panic(
  File "/workspaces/ZODB/src/ZODB/FileStorage/FileStorage.py", line 93, in panic
    raise CorruptedTransactionError(message % data)
ZODB.FileStorage.FileStorage.CorruptedTransactionError: oid 0x00 not found in prev_txn 0x03f6cffc5f37f1bb
```

What this test does is:
 - commits some transactions in "Source"
 - copy "Source" in "Dest"
 - pack "Dest"
 - undo a transaction in "Source"
 - restore the undo transaction (which has a `prev_txn`) in "Dest"

and this fails, because the prev_txn in "Dest" exists, but is a "p" (pack) transaction without records.

<details>
<summary>zodb dump Source.fs</summary>

```console
> zodb dump /tmp/FileStorageTestsao2yt_a3/FileStorageTestsy80r5b7k/FileStorageHexTestsemjrpbn0/FileStorageHexTestsyucvq6sd/FileStorageRecoveryTest9ur7kq7d/Source.fs
Ignoring index for /tmp/FileStorageTestsao2yt_a3/FileStorageTestsy80r5b7k/FileStorageHexTestsemjrpbn0/FileStorageHexTestsyucvq6sd/FileStorageRecoveryTest9ur7kq7d/Source.fs
txn 03f6d00826efb611 " "
user ""
description "initial database creation"
extension ""
obj 0000000000000000 64 sha1:d4b0bf30c5132b1ec57da7fe61b080970eea3418
�cpersistent.mapping
PersistentMapping
q.�}qXdataq}qs.

txn 03f6d00826f06aaa " "
user ""
description ""
extension ""
obj 0000000000000000 118 sha1:bd566fcc3ce469e0d6590f4b69d45e2d3708b4fa
�cpersistent.mapping
PersistentMapping
q.�}qXdataq}qXobj1qqcZODB.tests.MinPO
MinPO
q�qQss.
obj 0000000000000001 50 sha1:c69822b1b2dc00164c3db2db7d1013dcf706304b
�cZODB.tests.MinPO
MinPO
q.�}qXvalueqKs.

txn 03f6d00826f11488 " "
user ""
description ""
extension ""
obj 0000000000000000 148 sha1:4c39467361ff1ab627e82e14f9c4158cc36979f8
�cpersistent.mapping
PersistentMapping
q.�}qXdataq}q(Xobj1qqcZODB.tests.MinPO
MinPO
q�qQXobj2q      h�q
Qus.
obj 0000000000000002 50 sha1:c69822b1b2dc00164c3db2db7d1013dcf706304b
�cZODB.tests.MinPO
MinPO
q.�}qXvalueqKs.

txn 03f6d00826f4a100 " "
user ""
description "undo"
extension ""
obj 0000000000000000 from 03f6d00826f06aaa
obj 0000000000000002 delete

```
</details>


<details>
<summary>zodb dump Dest.fs</summary>

```console
> zodb dump  zodb dump /tmp/FileStorageTestsao2yt_a3/FileStorageTestsy80r5b7k/FileStorageHexTestsemjrpbn0/FileStorageHexTestsyucvq6sd/FileStorageRecoveryTest9ur7kq7d/Dest.fs
txn 03f6d00826f06aaa "p"
user ""
description ""
extension ""
obj 0000000000000001 50 sha1:c69822b1b2dc00164c3db2db7d1013dcf706304b
�cZODB.tests.MinPO
MinPO
q.�}qXvalueqKs.

txn 03f6d00826f11488 "p"
user ""
description ""
extension ""
obj 0000000000000000 148 sha1:4c39467361ff1ab627e82e14f9c4158cc36979f8
�cpersistent.mapping
PersistentMapping
q.�}qXdataq}q(Xobj1qqcZODB.tests.MinPO
MinPO
q�qQXobj2q      h�q
Qus.
obj 0000000000000002 50 sha1:c69822b1b2dc00164c3db2db7d1013dcf706304b
�cZODB.tests.MinPO
MinPO
q.�}qXvalueqKs.

```
</details>

I honestly don't fully understand what the test is doing with "copy the final transaction manually" and not using a higher level API, but this sounds like "this is more complex than just raising". Maybe this could be made a little bit more clever, tolerating the case where oid is not found because the transaction was packed, but for now, I believe we can merge this without trying to change to an error (or log or warning) - and then https://github.com/zopefoundation/ZODB/pull/395 .

Do you agree with this ? I will use "request review" and if you agree I'll apply the commits.

Thanks again for feedback.

d-maurer:
Jérome Perrin wrote at 2024-2-26 22:36 -0800:
> ...
>I believe we can merge this without trying to change to an error (or log or warning) - and then https://github.com/zopefoundation/ZODB/pull/395 .

I agree.

d-maurer (approved):

navytux:
Jérome, Dieter, thanks for feedback.

Personally I would still carefully reconsider turning `restore` inconsistencies into an error, but if you both think it is ok to go as it is please go ahead. I will be mostly away for the next 3 weeks and do not want to block progress. If you both think the patch is ok in its current form I trust you to make reasonable judgement even though personally I would try to do it differently.

Kirill

d-maurer:
Kirill Smelkov wrote at 2024-3-4 00:25 -0800:
>Personally I would still carefully reconsider turning `restore` inconsistencies into an error, but if you both think it is ok to go as it is please go ahead. I will be mostly away for the next 3 weeks and do not want to block progress. If you both think the patch is ok in its current form I trust you to make reasonable judgement even though personally I would try to do it differently.

As you know, I am less interested in potential problems with
low probability (than you). In may well be that a `restore` inconsistency
poses problems within some decades; I would then act -- with
a concrete case.

Unless you come with such a concrete case, I would not change
the current behavior - which apparently is check by a test case
(and thus likely carefully considered).

navytux:
Dieter, thanks for feedback.

The difference in between us is how we perceive importance of the bug here. For me a "data corruption" is the worst that could happen to a database. For an external observer data corruptions are usually hard to detect at the time when they happen, but they introduce a real change to the data. Some of such changes could be innocent, for example if we count a volume of some gas and it will be 17.3 litres instead of 17.2 that's quite tolerable and ok. Some of the changes are not so ok, for example if it will be 123 EUR in total after an exchange transaction instead of correct value of 321000 EUR. And some of the changes could really impact lives: for example if it will be the dose of radiation to emit, and due to corruption it will be 1000 Rentgens instead of 0.1 Rentgen, it makes a huge impact of the person in subject. There is *no* *way* to save life of that person with post-mortem analysis.

So when it comes to databases, I try to make dedicated attention to correctness first.

Still, as I said above, if both you and Jérome think it is ok to go as it is now, I trust you to make a reasonable choice and judgement.

Kirill

d-maurer:
Kirill Smelkov wrote at 2024-3-4 01:21 -0800:
> ...
>So when it comes to databases, I try to make dedicated attention to correctness first.
>
>Still, as I said above, if both you and Jérome think it is ok to go as it is now, I trust you to make a reasonable choice and judgement.

I read: you will come again in a few weeks with a concrete
case to reproduce the `restore` inconsistency.
Then, we will look again (in a separate PR).

perrinjerome:
Initially, this pull request was about extracting the +8 fix from https://github.com/zopefoundation/ZODB/pull/395 . I added testRestorePrevTxnWithoutOid , which is a scenario where the +8 bug would appear (because I don't like changing code without having a test corresponding to the change) but what does not seems good now is that this test would define the behavior for a case where the behavior was not clearly defined: what should happen when restore is called with a `prev_txn` which does not contain the `oid`.

Before all this, it is an undefined behavior, the data of the next transaction is read as a data header, which very likely causes an unexpected error to be raised. If the transaction is that last in the file, there is no error.

In the current version of the patch, it is tolerated and ignored.

From the discussions, @navytux suggested that this should cause a proper error.

I tried to change into an error, this caused a test to fail. The test is using low level APIs, so maybe it's just the test doing something wrong, I did not try to understand more.

Now I don't feel it would be good to merge testRestorePrevTxnWithoutOid as it is, because this would somehow be a specification of how FileStorage should behave when restore is called with such a transaction, but at this point I feel we don't really know what should be the behavior. The next step for this seems to be studying that failing test. I don't really have time to start doing this now.

https://github.com/zopefoundation/ZODB/pull/395 can be merged without all this, everybody approved so let's merge https://github.com/zopefoundation/ZODB/pull/395 already.

---

There was another commit in this pull request 7bef1b6e0ddc7bdfe4c1e680d099e8203f95956f , I think there was no controversy for that one.

navytux:
I reread this conversation carefully again. I think the reasonable outcome would be to:

- ignore `prev_txn` hint if the pointed-to transaction was packed,
- otherwise, if the pointed-to transaction is not packed, and does not contain oid - raise an error because this almost 100% surely represents invalid data passed to restore and relates to data corruption.

@perrinjerome, @d-maurer, do you agree with that path forward?

perrinjerome:
> I reread this conversation carefully again. I think the reasonable outcome would be to:
> 
> * ignore `prev_txn` hint if the pointed-to transaction was packed,
> * otherwise, if the pointed-to transaction is not packed, and does not contain oid - raise an error because this almost 100% surely represents invalid data passed to restore and relates to data corruption.

I also read everything again and that seems to make sense. For the "the pointed-to transaction is not packed, and does not contain oid" case, we can also treat in the same way the cases where the transaction does contains a record for oid, but with different data (different length/different content), these are also wrong cases.

I believe I understand the context, but I don't know the internals of ZODB very well, so I am not myself 100% confident with these changes, but the risk seems low, because this is only in the context of `restore`, which as far as I know is mostly used in https://github.com/zopefoundation/zc.zrs - and also I trust you on this. In other words, I believe that this plan is good, I can help on this topic, but I'm not really willing to push for this myself.

navytux:
@perrinjerome, thanks for feedback. Ok, I see. I will try to work on implementing this hopefully next week.

Reminder to myself: also raise an error if restore is invoked with wrong data (see https://github.com/zopefoundation/ZODB/pull/409#issuecomment-2788627507 for details).

PR #403, opened 2024-10-22
Repozo incremental recover

Sebatyne:
Like for the --backup mode of repozo, add an incremental feature for the recovery of a backup to a Data.fs, which allows to only append the latest backup(s) increment(s) (.deltafs) to a previously recovered Data.fs, instead of trashing it and restarted the recovery process from scratch.

This feature becomes the new default behavior, but the new flag "--force" allows to fall back to the previous behavior.

A few checks are done while trying to recover incrementally (ie: on size, or on the latest increment checksum), and the code automatically falls back to the full-recovery mode if they fail. This would happen for exemple if the production data has been packed after the previous recovery.

The need for such feature arose from our own production use, where we create delta backups of a file storage every day, send them to a stand-by server, and rebuild the ZODB there (still every day). When the ZODB is bigger than 1Tb, the full recovery can take several hours, whereas the incremental recovery would take a few minutes only (often even less)


Sebatyne:
@vpelletier , maybe you wish to review as you made the original work ?

@perrinjerome, you're the last contributor to repozo, could you review this PR ?

Thanks,

src/ZODB/scripts/repozo.py
```diff
@@ -755,6 +746,88 @@ def do_recover(options):
...
+            if endpos > initial_length:
+                break
+            previous_chunk = chunk
+        else:
+            if endpos == initial_length:
+                log('Target file is same size as latest backup, '
+                    'doing nothing.')
+                return
+            else:
+                log('Target file is longer than latest backup, '
```
perrinjerome, line 769: Is it correct to use "longer" / "shorter" and not "larger" / "smaller" ?

src/ZODB/scripts/repozo.py
```diff
@@ -755,6 +746,88 @@ def do_recover(options):
...
+                    'falling back to a full recover.')
+                return do_full_recover(options, repofiles)
```
perrinjerome, line 771: It seems the [test failure on windows](https://github.com/zopefoundation/ZODB/actions/runs/11452216534/job/31862720635#step:6:85) is because of here, we have opened `options.output` already and `do_full_recover` would re-open it a second time, which is not allowed on windows as far as I remember

perrinjerome (commented):
Can you also add a change log entry ?

I just looked at the diff this time, I will try to actually try running repozo with this patch later

Sebatyne:
Hello,

I would like to explain more my reasoning about the new behavior, as a change of default can be surprising.

A good practice with a backup/recovery plan is to check that the backed-up data can be restored on a remote service. That's why we recover the Delta.fs every day, to check that the latest .deltafs increment (which is the only new backed-up file every day, as the other .deltafs and the original .fs are already synchronised on the remote site) is valid.

From this observation, as when we import the new increment, we already have the recovered Delta.fs from the previous day, it sounds a waste of resource to delete it, and rebuild it from 0. If we could simply recover the new increment on the existing Delta.fs, then its sum would be checked, proving its validity once and for all. And we don't need to check its validity every day, as a data corruption is most likely to happen during the write process or the network copy.

Also, I believe the time saved to not restore a full Data.fs is welcome, as it allows to decrease the time-to-recovery in case of activation of the disaster recovery plan, or simply to create backups more often, to decrease the quantity of lost data in a production incident.

Please feel free to ask me more questions.

Regards,

Nicolas

src/ZODB/scripts/repozo.py
Sebatyne: Thanks, I edited to use "larger" / "smaller". I think this comes from the calls to .seek() and .tell(), which makes the file looks like a long line, and not like an object.

Sebatyne:
> Can you also add a change log entry ?
> 
> I just looked at the diff this time, I will try to actually try running repozo with this patch later

I have added an entry. But I'm not sure about the wording.

mgedmin (changes requested):
Overall I like this a lot.  I've two small fixes to suggest.

src/ZODB/scripts/repozo.py
```diff
@@ -755,6 +746,92 @@ def do_recover(options):
             raise
 
 
+def do_incremental_recover(options, repofiles):
+    datfile = os.path.splitext(repofiles[0])[0] + '.dat'
+    log('Recovering (incrementally) file to %s', options.output)
+    with open(options.output, 'r+b') as outfp:
+        outfp.seek(0, 2)
+        initial_length = outfp.tell()
+    with open(datfile) as fp:
```
vpelletier, line 755: Doesn't reopening the index risk disobeying `find_files` logic ? Especially, the `--date` option.

Sebatyne:
Sorry for the long list of "fixup!" commits, I didn't think it would get that long...

To implement the feedback received in this MR, and to prevent an error on windows because a same file was opened twice, I had to rework deeply the function `do_incremental_recover`. I hope it is not (too much...) an issue for the review.

I have added more assertions in acadc7a5697cf8a68769668bd7ad84c6cbbbe01c, as well as a new step where I delete an old .deltafs already recovered, to prove the correctness of the code, and that it doesn't fall back silently to the full-recovery mode. I hope it will help you trust the rewriting of `do_incremental_recover` that happened in the latest commits.

Sebatyne:
Hello,

Sorry, it's 2 months later, and I just found out about "pending" comments in github, and how to submit them using the hidden button (it seems I'm not the only one to have been confused: https://github.com/orgs/community/discussions/10369)

So, a summary is much needed I believe...

I have implemented everyone's recommendations.

During this PR, a big change was the inclusion of a "--quick" flag for the incremental recovery. In my first implementation the code would only check the latest chunk of the already recovered data.fs, and then decide if it would go for an incremental recovery (the new default), or a full recovery. Based on @perrinjerome's and @vpelletier's advices, the default would be to check all the known chunks from the .dat file against the already recovered data.fs  instead of only the latest. But the "check only the latest" behavior still exists, behind the "--quick" flag.

I have merged all my commits and I believe this PR is ready to merge.

I'm sorry again if I gave the impression to abandon this PR and ignoring your comments. This is not the case, and it is due to my clumsiness with the github interface...

Regards,

perrinjerome (approved):
I tried again the current version of the patch and it seems fine. @mgedmin can you take another look now that what you raised have been addressed ?

Sebatyne:
Hello @mgedmin .  I see you didn't give news for a long time, and I hope you are doing fine.

I consider this MR ready for the merge, but I cannot do this until you give your approval. Could you review the 2 points you raised, and approve this MR ? Or give me more work to do ;)

Regards,

mgedmin:
I'm sorry, the burnout is too strong, I was unable to force myself to look at the latest diff.

navytux:
@mgedmin, thanks a lot for feedback. That's sad to hear. I understand the times are difficult, wish you to keep safe.

PR #404, opened 2024-10-29
repozo: prevent an incorrect "option ignored" warning

perrinjerome:
When running backup or verify, "--with-verify option is ignored" warning was issued even when this option was not passed


Sebatyne (approved):

perrinjerome:
Thanks !

PR #405, opened 2024-10-29
repozo: fix recover on stdout

perrinjerome:
This was working on python2, but on python3,

```
  repozo -v --recover --repository backups/ > restored.fs
```

failed with

```  
  TypeError: write() argument must be str, not bytes
```


perrinjerome:
I have no idea how we can test this. 

It seems a minor problem, I just noticed the error by forgetting the `-o` argument

PR #409, opened 2025-04-07
FileStorage: Fix restore to correctly check restored data content

navytux:
In commit 8160568b (FileStorage: fix rare data corruption when using
restore after multiple undos (#395)) we fixed FileStorage.restore to
work correctly in the presence of multiple undos for the same oid in the 
same transaction by reworking oid data records of one transaction to be
all scanned in the loop. But unfortunately that commit introduced a
regression:

When restore is instructed to do an undo, in other words, to make a copy
of oid data by referring to that data by backpointer, it checks content
of the pointed-to record, and if data content there does not match data
content passed to restore call, the copy-via-backpointer optimization is
silently disabled. However 8160568b contained a thinko when doing the 
check: after discovering pointed-to oid data record, the intent there it
was to load actual data from there, but the loading was performed not 
from the offset where the data lives, but from then beginning of some
next data-record header instead. As the result the 
"copy-via-backpointer" was always disabled.

-> Fix that by seeking to correct position before loading the data.

I originally discovered this issue by running Zodbtools tests wrt ZODB6
and getting failures for `zodb restore` related tests with difference in
between restored and expected states as e.g.

    @@ -296,15 +296,21 @@
     txn 0285cbad77777800 " " 
     user "root1.0\nYour\nMagesty "
     description "undo 1.0\nmore detailed description\n\nzzz ...\t"
     extension ""
    -obj 0000000000000005 from 0285cbad6962fd19
    +obj 0000000000000005 30 sha1:5fe26f63145e7ad5ac7c2a6bc813e706c5548278
    +c__main__
    +Object
    +q^A.U^Ea1.22q^B.

/cc @perrinjerome, @d-maurer

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,12 @@
 
  - repozo: prevent an incorrect "option ignored" warning when running backup or verify.
 
+ - FileStorage: fix `restore` regression introduced in ZODB 6.0 in `#395
+   <https://github.com/zopefoundation/ZODB/pull/395>`_: when restoring data
+   records with undo the `restore` was no longer emitting backpointers and was
+   emitting duplicate data copies instead. `#409 <https://github.com/zopefoundation/ZODB/pull/409>`_
+   fixes `restore` back to emit data records with backpointers for undo again.
+
 
 6.0 (2024-03-20)
 ================
--- a/src/ZODB/FileStorage/FileStorage.py
+++ b/src/ZODB/FileStorage/FileStorage.py
@@ -704,6 +704,7 @@ class FileStorage(
                 logger.error("Mismatch between data and"
                              " backpointer at %d", pos)
                 return 0
+            self._file.seek(data_pos + DATA_HDR_LEN)
             _data = self._file.read(data_hdr.plen)
             if data != _data:
                 return 0
--- a/src/ZODB/tests/testFileStorage.py
+++ b/src/ZODB/tests/testFileStorage.py
@@ -35,6 +35,7 @@ from ZODB._compat import dump
 from ZODB._compat import dumps
 from ZODB.Connection import TransactionMetaData
 from ZODB.fsIndex import fsIndex
+from ZODB.interfaces import IStorageWrapper
 from ZODB.tests import BasicStorage
 from ZODB.tests import ConflictResolution
 from ZODB.tests import Corruption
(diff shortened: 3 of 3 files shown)


navytux:
(rebased after https://github.com/zopefoundation/ZODB/pull/408  merge)

perrinjerome:
I'll look at this hopefully next week. I read the description and it makes sense, but I don't understand the code changes, I need to take time to remember how this works. This fixes the failing zodbtools test and still passes all ZODB tests, so it's probably good.

From my understanding, the problem was that when restoring an undo, we are supposed to store a pointer to the previous record, but the introduced regression is that we no longer store a pointer but a full copy of the data record, so the problem has a low impact, the data is correct, it's just that the optimization to just store a back-pointer did not happen, so loading the record data is correct, but the restored storage is not byte to byte equivalent. I think it would be good to have one line in the change log, at least to clarify the impact of the problem and because it's common to have change logs in zopefoundation repositories.

navytux:
Jérome, thanks for feedback.

Yes, your understanding is correct - restore was buggy when checking the original data content and because of that bug was practically disabling "copy via backpointer" optimization. The data was restored as full copy in that case. Indeed we were lucky to not get a corruption, but database restoration is no longer bit-to-bit identical to the original. And "Undo history" also became broken due to that bug. This still can be viewed as data bug and is important to fix.

Thanks for your suggestion about changelog. I've amended the PR correspondingly. 

Looking forward for your further feedback.

Kirill

perrinjerome:
Thanks, I read all this again and I understand the problem now and this is a correct fix.

> By the way - from the perspective of https://github.com/zopefoundation/ZODB/pull/397 (see summary in https://github.com/zopefoundation/ZODB/pull/397#issuecomment-2782898179) I suggest that we also do ( ... )

Ah yes, it's very similar to the `logger.error("Mismatch between data and"` that is logged when `prev_txn` contains a record for oid but the data have different lengths. When it's the same length but different data, nothing is logged, but it's also a case of "Something is wrong".

perrinjerome (approved):

navytux:
@perrinjerome, thanks for feedback and for the approval. I think it is ok to merge this patch. @d-maurer, please speak up if you have something to add/correct, but hopefully it should be ok.

Regarding the "something wrong" let's continue the discussion in https://github.com/zopefoundation/ZODB/pull/397.

navytux:
@dataflake, as this patch fixed a regression introduced in ZODB 6.0, do you think we can make a new 6.0.1 release with the merged fix?

Thanks beforehand,
Kirill

dataflake:
6.0.1 is now released.

navytux:
Thanks, @dataflake.

PR #415, opened 2025-09-29
Apply the latest zope.meta templates

dataflake:
This PR brings the current ZODB master branch up to date with the standards we are applying across all zopefoundation packages in terms of supported Python versions and package configuration files.

I realize this is a large PR that contains a lot of cosmetic changes, but since no one has touched ZODB in quite a while that kind of bit rot was bound to happen. There is no point in letting it fall back even further.


src/ZODB/scripts/repozo.py
```diff
@@ -788,7 +788,7 @@ def do_verify(options):
             elif not options.quick:
                 if actual_sum != sum:
                     raise VerificationFail(
-                        "%s has checksum %s%s instead of %s" % (
+                        "{} has checksum {}{} instead of {}".format(
                             filename, actual_sum, when_uncompressed, sum))
```
mauritsvanrees: [Edited: sorry, I pressed the wrong button, I wanted to edit this still.
I can't suggest an edit that includes a line that did not changed.]

We should use f-strings here as well.
 
```
raise VerificationFail(
    f"{filename} has checksum {actual_sum}{when_uncompressed} instead of {sum}")
```

src/ZODB/scripts/migrate.py
```diff
@@ -310,8 +310,23 @@ def doit(srcdb, dstdb, options):
...
+             t4 -
+             t0,
+             t1 -
+             t0,
+             t2 -
+             t1,
+             t3 -
+             t2,
+             t4 -
+             t3),
```
mauritsvanrees: This automatic change looks ugly.  When I edit it like this locally, pre-commit is still happy:

```
             t4 - t0,
             t1 - t0,
             t2 - t1,
             t3 - t2,
             t4 - t3),
```

(It looks like I can't get a suggestion on a multiline change to work, so just pasted it as code here.)

src/ZODB/scripts/repozo.py
dataflake: Can you please use the suggestion functionality so I don't have to type this change in manually?

mauritsvanrees (approved):
Looks good, thanks!
I tried it in the Plone 6.1 core development buildout on Python 3.13, and that was fine.

I made two small suggestions, but they need not hold back a merge, so I approve.

src/ZODB/scripts/repozo.py
mauritsvanrees: That is what I meant: I tried that, but it failed.

dataflake: Sorry, I misunderstood. Than you for the suggestion and review!

src/ZODB/scripts/migrate.py
dataflake: Done, thanks!

docs/articles/old-guide/chatter.py
```diff
@@ -81,20 +84,20 @@ def get_chat_session(conn, channelname):
     # the key 'chat_sessions'.
     root = conn.root()
     if not root.has_key('chat_sessions'):
-        print 'Creating chat_sessions B-tree'
+        print 'Creating chat_sessions B-tree'  # NOQA: E999 print statement
```
icemac:
```suggestion
        print('Creating chat_sessions B-tree')
```

This will even work in Python 2.7.

docs/articles/old-guide/convert_zodb_guide.py
```diff
@@ -1,18 +1,19 @@
...
+from converter import restwriter
 
 
 if __name__ == '__main__':
     try:
         rootdir = sys.argv[1]
         destdir = os.path.abspath(sys.argv[2])
     except IndexError:
-        print "usage: convert.py docrootdir destdir"
+        print "usage: convert.py docrootdir destdir"  # NOQA: E999 print
```
icemac:
```suggestion
        print("usage: convert.py docrootdir destdir")
```

icemac (approved):
LGTM scrolling through the changes expecting that nearly all of them are automatically generated.

dataflake:
The changes are all auto-generated, correct. I used the NOQA instead of making code changes because the code in question was already old, obsolete and unused. I did apply your suggestions, though, thanks!

dataflake:
I have just released ZODB version 6.1.

navytux:
@dataflake, I'm late to the party but thanks for caring about maintenance.

dataflake:
@navytux If you have any active branches I would volunteer to merge master.

PR #416, opened 2026-01-23
Replace pkg_resources by importlib.metadata.

icemac:


dataflake:
I have created https://github.com/zopefoundation/ZODB/pull/417 to use the latest zope.meta templates, which will fix the test failure here.

dataflake (approved):

PR #423, opened 2026-09-14
Update to new pre-commit hooks.

icemac:
- **Update to new pre-commit hooks.**
- **Pre-commit auto-fix.**
- **Manual fixes.**


dataflake (changes requested):
You should exclude `docs/zodb.svg` and `src/ZODB/scripts/manual_tests/test-checker.fs` from processing

icemac:
Done in 0c03a98d: reverted the whitespace changes to `docs/zodb.svg` and `src/ZODB/scripts/manual_tests/test-checker.fs` and excluded both from `end-of-file-fixer` and `trailing-whitespace` (via `whitespace-exclude` in `.meta.toml`, so it survives regeneration).

— _Comment created by Claude_

dataflake (approved):

